PR #10, opened 2024-01-23
DFAST-49: update gui and user manual

hrajagers:
* Update the GUI (not yet fully functional) to version 2 files for river and analysis configurations.
* Update the user manual to match the new input files and GUI; the CLI mode is now moved to the appendix.
* Add a new pre-draft version of the Dutch_rivers.ini file (final version to be agree upon with RWS) ... the old file renamed to Dutch_rivers_v1.ini such that the CLI and BATCH modes can continue to use them for repeating old analyses.
* Update the old test cases: they all use the old Dutch_rivers_v1.ini configuration.
* Two new basic distribution tests added to check whether the "dfastmi.exe --help" runs, and that it produces the expected result.


hrajagers:
@MRVermeulenDeltares, @ralph-peelen-deltares - Can you have a look at this pull request?

dfastmi/__main__.py
```diff
@@ -113,7 +113,7 @@ def parse_arguments() -> Tuple[str, str, Optional[str], str, bool]:
         "--rivers",
         help="name of rivers configuration file ('Dutch_rivers.ini' is default)",
     )
-    parser.set_defaults(rivers="Dutch_rivers.ini")
+    parser.set_defaults(rivers="unspecified")
```
MRVermeulenDeltares, line 116: Is there a specific reason to put it to unspecified, since lateron the riverfile will be put to a file based on the runmode.
Wouldn't it be better to use that if here instead of using the "unspecified"?
e.g.:

`if runmode == "CLI":
            parser.set_defaults(rivers="Dutch_rivers_v1.ini")"
        else:
            parser.set_defaults(rivers="Dutch_rivers.ini")`

dfastmi/gui.py
```diff
@@ -604,7 +650,7 @@ def main(rivers_configuration: RiversObject, config: Optional[str] = None) -> No
     rivers = rivers_configuration
     create_dialog()
     dialog["branch"].addItems(rivers["branches"])
-    dialog["reach"].addItems(rivers["reaches"][0])
+    #dialog["reach"].addItems(rivers["reaches"][0])
```
MRVermeulenDeltares, line 653: Please remove this line if it is not used.

dfastmi/messages.UK.ini
```diff
@@ -293,46 +293,18 @@ On which river reach is the measure located?
 Discharge location
```
MRVermeulenDeltares, line 293: Should these changes also be applied to message.NL.ini?

MRVermeulenDeltares: Are some of these actually used in the software?

tests-dist/test_basic.py
```diff
@@ -0,0 +1,71 @@
...
+        cwd = os.getcwd()
+        tstdir = "tests/c01 - GendtseWaardNevengeul"
+        try:
+            os.chdir(tstdir)
+            result = subprocess.run([dfastexe,"--help"], capture_output=True)
+            outstr = result.stdout.decode('UTF-8').splitlines()
+        finally:
+            os.chdir(cwd)
+        #
+        #for s in outstr:
```
MRVermeulenDeltares, line 53: please remove commented code

tests/test_batch.py
```diff
@@ -26,7 +26,7 @@ def test_batch_mode_00(self):
         Testing batch_mode: missing configuration file.
```
MRVermeulenDeltares, line 26: there are no longer any tests for dutch_rivers in this case?

MRVermeulenDeltares:
@hrajagers Reviewed the changes.
I am unsure how extensive I should have reviewed, currentlty I kept it at a light review, since future refactors will improve it more.
I did not review the .tex files.
I will discuss my review with Ralph and come back to you after that.

dfastmi/__main__.py
hrajagers: The reason is that the `runmode` is not yet known ... that is one of the arguments to be parsed. Basically, we check whether the `rivers` argument has been specified or not. If not, then we use the default value which depends on the `runmode` (_v1 for the backward compatible CLI mode, and normal version ... maybe we should rebrand it to _v2 for the BATCH and GUI mode)

dfastmi/gui.py
hrajagers: Done

dfastmi/messages.UK.ini
hrajagers: Yes, but while experimenting with the GUI layout, I didn't want to each time create and remove entries in both language files. I've now updated the message.NL.ini file as well.

They should all be used in the software ... there is currently no check on that ... You can find the places where these strings are used by searching for `dfastmi.io.log_text` calls in the text (all keys not starting with "gui_") and for `gui_text` text calls (which automatically adds the prefix "gui_" for recognizability)

tests-dist/test_basic.py
hrajagers: Done.

tests/test_batch.py
hrajagers: Correct. Or phrased differently: the are no tests YET for the new Dutch_rivers.ini file. All existing tests still exist, but they are all based on the old Dutch_rivers_v1.ini configuration (which was previously called Dutch_rivers.ini). We'll start adding new cases using new functionality based on the new configuration file, but the new configuration file isn't final yet. The content will change significantly and almost all possible tests using it would have to get new references when it's updated to its final form.

hrajagers: Note: I've now renamed the new Dutch_rivers.ini file to Dutch_rivers_v2.ini to make it more clear that this is a new file. The content is still subject to change. We can use a simpler file for basic unit testing.

dfastmi/__main__.py
hrajagers: Note: I've now renamed the new Dutch_rivers.ini file to Dutch_rivers_v2.ini to make it more clear that this is a new file.

hrajagers:
The SonarCloud code coverage percentage is relatively low for these changes since most changes were located in the GUI code and that part of the code is not yet covered by automated tests.

ralph-peelen-deltares:
Latex files are not reviewed, we prefer them in a separate issue next time. When we release we should validate the manuals.

PR #14, opened 2024-01-26
DFAST-83: Update nuitka build config with include-data-files

ralph-peelen-deltares:
- Removed the unneeded scripts

### DIFF
--- a/BuildScripts/BuildDfastmi.bat
+++ b/BuildScripts/BuildDfastmi.bat
@@ -17,9 +17,11 @@ START /B /WAIT poetry run nuitka ^
  --include-module=geopandas ^
  --include-package-data=geopandas.datasets ^
  --include-module=fiona ^
+ --include-data-files=dfastmi/Dutch_rivers.ini=dfastmi/Dutch_rivers.ini ^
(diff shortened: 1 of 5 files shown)


MRVermeulenDeltares (approved):
Changes might get into conflic with https://github.com/Deltares/D-FAST_Morphological_Impact/pull/10

hrajagers:
> Changes might get into conflic with #10

Yes, they did. I've merged the changes of this branch into #10 and resolved the conflict.

hrajagers:
NOTE: The Nuitka command in BuildDfastmi.bat now includes " --include-data-files=docs/dfastmi_usermanual.pdf=dfastmi/dfastmi_usermanual.pdf ^". I don't think that that command will work since the pdf is not included in the source code. It is only generated by the Build process on TeamCity ... most likely on another node. That's why we merge the pdf into the distribution as part of the TeamCity workflow.

MRVermeulenDeltares:
This works, since the Nuitka build on teamcity is dependant on the latex generation build.
The pdf will be available before the Nuitka build on the agent starts, thus is will be included in the .exe

PR #19, opened 2024-02-05
DFAST-89 investigate and adapt new structure for kernel

MRVermeulenDeltares:


dfastmi/kernel/typehints.py
```diff
@@ -0,0 +1,7 @@
+from typing import Tuple, Optional
+
+QLevels = Tuple[float, float, float, float]
```
priscavdsluis, line 3: Perhaps try to document the values within the tuples

MRVermeulenDeltares: Done

ralph-peelen-deltares (approved):

PR #20, opened 2024-02-05
DFAST-92 Refactor the main_computation method in Kernel into the BedLevelCalculator class

MRVermeulenDeltares:
This PR is focused on moving the main_computation into BedLevelCalculator with some refactoring.
This PR is dependant of PR #19


dfastmi/kernel/BedLevelCalculator.py
```diff
@@ -0,0 +1,144 @@
...
+import numpy
+from dfastmi.kernel.typehints import Vector
+
+
+class BedLevelCalculator:
+    """
+    This class is used to calculate bed level related values.
+    """
+
+    def __init__(self, number_of_periods: int):
```
tim-vd-aardweg, line 11: In python you can specify that you expect `number_of_periods` to be an `int`, however, in python it is merely a suggestion :D

If you want to fail fast, you could add a check if the `number_of_periods` is indeed an `int`. Currently, if a non-integer is provided, it will probably crash at a later point with potentially a more vague error message.

dfastmi/kernel/BedLevelCalculator.py
```diff
@@ -0,0 +1,144 @@
...
+            List of arrays containing the bed level change at the beginning of each respective discharge period.
+
+        Returns
+        -------
+        dzmin : numpy.ndarray
+            minimum bed level change.
+        """
+        return numpy.minimum.reduce(dzb)
+
+    def get_linear_average(self, number_of_days: Vector, dzb: List[numpy.ndarray]) -> numpy.ndarray:
```
tim-vd-aardweg, line 50: - This method can raise an Error if the length of `number_of_days` or `dzb` is less than `number_of_periods`.
- I know we talked about it shortly, but I am a little confused about the `number_of_days`. To me, it sounds as if it should be an integer (if I only look at the name), but it is a `Vector`.

dfastmi/kernel/BedLevelCalculator.py
```diff
@@ -0,0 +1,144 @@
...
+            A list of arrays containing the equilibrium bed level change for each respective discharge period.
+        rsigma : Vector
+            A tuple of relaxation factors, one for each period.
+
+        Returns
+        -------
+        dzb   : List[numpy.ndarray]
+            List of arrays containing the bed level change at the beginning of each respective discharge period.
+        """
+        vsigma = self.__compute_vsigma__(rsigma, dzq)
```
tim-vd-aardweg, line 89: I think you should not end the method name with __. Now it suggests this is a 'dunder' method, while by the looks of it, it is just a 'private' method.

dfastmi/kernel/BedLevelCalculator.py
```diff
@@ -0,0 +1,144 @@
...
+        den = self.__compute_denominator__(self.number_of_periods, vsigma)
+        dzb = self.__compute_dzb_at_the_beginning_of_each_period__(dzq, vsigma, den)
+        return dzb
+
+    def __compute_vsigma__(self, rsigma : List[float], dzq):
```
tim-vd-aardweg, line 94: What is the type of `dzq`?

dfastmi/kernel/BedLevelCalculator.py
```diff
@@ -0,0 +1,144 @@
...
+        This routine computes the bed level changes.
+        This routine requires that dzq and rsigma have the same length.
+        A stagnant period can be represented by a period with rsigma = 1.
+
+        Arguments
+        ---------
+        dzq : List[numpy.ndarray]
+            A list of arrays containing the equilibrium bed level change for each respective discharge period.
+        rsigma : Vector
+            A tuple of relaxation factors, one for each period.
```
tim-vd-aardweg, line 82: The typehint says it's a `Vector` but in the description it is just a tuple?

dfastmi/kernel/BedLevelCalculator.py
tim-vd-aardweg: Same for the other methods ending in __ ofcourse ^^

dfastmi/kernel/core.py
```diff
@@ -309,69 +310,13 @@ def main_computation(
...
-            dzmin = numpy.minimum(dzmin, dzb[i])
-
-    # linear average
-    for i in range(N):
-        if i == 0:
-            dzgem = dzb[0] * (T[0] + T[-1]) / 2
-        else:
-            dzgem = dzgem + dzb[i] * (T[i] + T[i-1]) / 2
+    """  
+    number_of_periods = len(dzq)
```
tim-vd-aardweg, line 314: Nice and clean 👍

dfastmi/kernel/core.py
```diff
@@ -282,7 +283,7 @@ def dzq_from_du_and_h(
 
 def main_computation(
     dzq: List[numpy.ndarray],
-    T: Vector,
+    number_of_days: Vector,
     rsigma: Vector,
 ) -> Tuple[numpy.ndarray, numpy.ndarray, numpy.ndarray, List[numpy.ndarray]]:
```
tim-vd-aardweg, line 288: Can we do something about the return type? What does the `Tuple[numpy.ndarray, numpy.ndarray, numpy.ndarray, List[numpy.ndarray]]` represent? :D

dfastmi/kernel/BedLevelCalculator.py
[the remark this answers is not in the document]
MRVermeulenDeltares: I wasnt sure what to do for the conventions and was in discussion with Ralph how to do this.
It might be good if we have a standard (Deltares) python convention an apply it on the python projects we work on.
E.g. dfast, singlerunner, hydrolib, etc..
I created this issue to tackle it for dfast.
Maybe we can use a package or tool to help us with that, https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html or so.

dfastmi/kernel/BedLevelCalculator.py
MRVermeulenDeltares: That is currently what I am doing with typehinting right?

Any suggestion on how to tackle the second part?

dfastmi/kernel/BedLevelCalculator.py
```diff
@@ -0,0 +1,144 @@
+from typing import List
+import numpy
```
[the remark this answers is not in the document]
MRVermeulenDeltares: I moved the import from the core.py file to this one, I can adjust it to your suggestion.

dfastmi/kernel/BedLevelCalculator.py
MRVermeulenDeltares: Ah I though __name__ signified that it is a private method, this is not the case?
A private method would be: __name?

dfastmi/kernel/BedLevelCalculator.py
MRVermeulenDeltares: List[numpy.ndarray], but since it is a private method I felt that typehinting was less needed since this should already be passed into the parent method correctly.
The rsigma is typehinted because a ceirtain List functionality is needed (I think because of append)

dfastmi/kernel/BedLevelCalculator.py
MRVermeulenDeltares: The Vector and BoolVector are more arrays, but if you check the typehints.py you can see that they are currently stated as tuple[float,...]

dfastmi/kernel/BedLevelCalculator.py
MRVermeulenDeltares: To answer the first part:
in the original main_computation you could find the following: 

```
    N = len(dzq)
    # N should also equal len(T) and len(rsigma)
```
From this I derived the constructor of this class and changed N into self.number_of_periods.
The "number_of_days" was previously "T", which is a vector and in turn a tuple[float,...].
The dzb is derived from N, and should be the same/similar size as dzq I think.

I wasnt sure up to what point I should have been putting validation on everything put into a method.
I can add the suggestion to the documentation.

MRVermeulenDeltares: To answer the second part.
The "number_of_days" was previously "T", which is a vector and in turn a tuple[float,...].
I based the name on the documentation which was available regarding "T", it could be the name "number_of_days" isn't correct or is lacking.

```
    T : Vector
        A tuple of periods indicating the number of days during which each discharge applies.
```

I think one float in the tuple represents a 24h day, e.g. 0.5 would be 12h.
But I am not sure about this.
I made a comment in regard to the vector/tuple[float,...] in a review comment below.

dfastmi/kernel/core.py
```diff
@@ -309,69 +310,13 @@ def main_computation(
         Minimum bed level change.
     dzb   : List[numpy.ndarray]
         List of arrays containing the bed level change at the beginning of each respective discharge period.
-    """
```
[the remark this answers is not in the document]
MRVermeulenDeltares: yes, see "class Test_main_computation():" in:
https://github.com/Deltares/D-FAST_Morphological_Impact/blob/master/tests/kernel/test_core.py

dfastmi/kernel/core.py
MRVermeulenDeltares: dzgem, dzmax, dzmin, dzb.
Nice and clean Single responsibility of the main_computation method xD

I wanted to tackle that in a seperate issue, since the 4 calls from the bedlevelcalculator can replace this method and seperate the 4 returns. (https://issuetracker.deltares.nl/browse/DFAST-93)
After discussion with Prisca we decided to leave it in main_computation for now.
This can still be tackled in the future, since batch.py also needs refactoring. (the user or this method)

Besides this, I can still put the above mentioned return type as a typehint if that would help with readability, but I do not think it will help much...

tests/kernel/test_BedLevelCalculator.py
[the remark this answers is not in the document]
MRVermeulenDeltares: see comment above regarding conventions

dfastmi/kernel/BedLevelCalculator.py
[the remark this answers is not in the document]
MRVermeulenDeltares: Discussed, @MRVermeulenDeltares will plan something with the Python developers of the Purple team.

dfastmi/kernel/BedLevelCalculator.py
MRVermeulenDeltares: discussed: assert (maybe?) or if raise when number_of_periods is not an int.

dfastmi/kernel/BedLevelCalculator.py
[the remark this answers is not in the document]
MRVermeulenDeltares: Will be discussed in DFAST-101

dfastmi/kernel/BedLevelCalculator.py
MRVermeulenDeltares: Discussed, first part will be discussed in dfast-101.
second part, ask Bert if the can explain more what T is, to get a better name for that.

dfastmi/kernel/BedLevelCalculator.py
MRVermeulenDeltares: Discussed, @MRVermeulenDeltares will adjust

dfastmi/kernel/BedLevelCalculator.py
MRVermeulenDeltares: discussed: Will be further discussed in dfast-101, can be update for this now.

dfastmi/kernel/BedLevelCalculator.py
MRVermeulenDeltares: discussed, no changes, ready to resolve.

tests/kernel/test_BedLevelCalculator.py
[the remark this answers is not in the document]
MRVermeulenDeltares: will be tackled in 101

dfastmi/kernel/core.py
MRVermeulenDeltares: discussed, resolve conversation.

dfastmi/kernel/core.py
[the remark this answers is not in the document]
MRVermeulenDeltares: discussed, resolve conversation.

dfastmi/kernel/BedLevelCalculator.py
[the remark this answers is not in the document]
MRVermeulenDeltares: Planned monday the 19th, I will keep updating the 101 issue before the meeting if I find new things.

dfastmi/kernel/BedLevelCalculator.py
MRVermeulenDeltares: added

dfastmi/kernel/BedLevelCalculator.py
MRVermeulenDeltares: done

dfastmi/kernel/BedLevelCalculator.py
MRVermeulenDeltares: done

dfastmi/kernel/BedLevelCalculator.py
MRVermeulenDeltares: I will ask if Bert can review it based on variable naming

tim-vd-aardweg (approved):

dfastmi/kernel/BedLevelCalculator.py
```diff
@@ -0,0 +1,149 @@
...
+        -------
+        dzb   : List[numpy.ndarray]
+            List of arrays containing the bed level change at the beginning of each respective discharge period.
+        """
+        vsigma = self.__compute_vsigma(rsigma, dzq)
+        den = self.__compute_denominator(self.number_of_periods, vsigma)
+        dzb = self.__compute_dzb_at_the_beginning_of_each_period(dzq, vsigma, den)
+        return dzb
+
+    def __compute_vsigma(self, rsigma : List[float], dzq: List[numpy.ndarray]) -> List[numpy.ndarray]:
```
hrajagers, line 99: Should we add descriptions here too? Or is that only needed for the public functions?
Furthermore, the `List[float]` type is equivalent to `Vector` in the other routines.
Yes, the lists are indeed frequently referred to as tuples in the comments. I've used tuples before since that was better for  the static type checker "mypy" and all tuples had exactly length 3. Now, they can have any length and lists are easier for that ...

dfastmi/kernel/BedLevelCalculator.py
```diff
@@ -0,0 +1,149 @@
...
+    
+    def __get_mask(self, dzq: List[numpy.ndarray]) -> numpy.ndarray[bool]:
+        mask = numpy.zeros_like(self.number_of_periods, dtype=bool)
+        for dzq_value in dzq:
+            if dzq_value is not None:
+                mask = mask | numpy.isnan(dzq_value)
+        return mask
+
+    def __compute_denominator(self, number_of_periods: int, vsigma: List[numpy.ndarray]) -> int:
+        for i in range(number_of_periods):
```
hrajagers, line 119: Why not use self.number_of_periods?

hrajagers (commented):
I've not checked the code that has been moved around. I've focused on the names used. Sorry for the mismatch between old documentation and actual meaning. That's why we need this clean-up phase.

dfastmi/kernel/BedLevelCalculator.py
MRVermeulenDeltares: I will update this.

dfastmi/kernel/BedLevelCalculator.py
MRVermeulenDeltares: "Private" method do not necessarily need documentation, since they are "private" and no user of the class or module should use it.

I will change the vector and boolvector from a tuple to a list.
I will change the List[float] in this class to vector and update documentation.

tim-vd-aardweg (approved):

PR #22, opened 2024-02-07
Feature/dfast 105 add a first testcase using version 2 configuration files

hrajagers:
Add a test case for version 2 configuration files.
Fix one issue with version 2 configuration files: it should be allowed to not specify the "Tide" keyword in which case it should default to "False".


tests/test_batch.py
```diff
@@ -157,3 +158,44 @@ def test_batch_mode_03(self):
...
+                dfastmi.batch.batch_mode("Qmin_4000_v2.cfg", rivers, False)
+            outstr = out.getvalue().splitlines()
+        finally:
+            os.chdir(cwd)
+        #
+        #for s in outstr:
+        #    print(s)
+        self.maxDiff = None
+        print(outstr)
+        #assert outstr == []
```
MRVermeulenDeltares, line 184: should this be commented?

MRVermeulenDeltares: fixed

MRVermeulenDeltares (approved):

PR #23, opened 2024-02-08
DFAST-95: Refactor estimate_sedimentation_length method

MRVermeulenDeltares:


dfastmi/kernel/core.py
```diff
@@ -244,10 +212,9 @@ def estimate_sedimentation_length2(
     L : float
         The expected yearly impacted sedimentation length [m].
     """
-    Lt = [tmi[i] * celerity[i] for i in range(len(tmi))]
+    yearly_sedimentation_lengths = [tmi[i] * celerity[i] for i in range(len(tmi))]
```
hrajagers, line 215: The term `yearly_sedimentation_lengths` isn't appropriate. The variable represents the contribution of each period/discharge level to the total sedimentation length, so let's use `sedimentation_length_contributions` instead.

dfastmi/kernel/core.py
```diff
@@ -244,10 +212,9 @@ def estimate_sedimentation_length2(
...
     
-    return sum(Lt) * 1000
-
+    return sum(yearly_sedimentation_lengths) * 1000
```
hrajagers, line 217: Let's introduce a constant `km_to_m` equal to 1000 to show where it's coming from.

hrajagers (changes requested):
The moving of code around looks fine. Just two small comments about the improvements to `core.estimate_sedimentation_length`.

hrajagers:
Hi @MRVermeulenDeltares, I removed the call to kernel.legacy.sedimentation_length ... so, you can remove that legacy function and the associated tests.

I've left my recommendations for the newer kernel.core.sedimentation_length to you.

dfastmi/kernel/core.py
MRVermeulenDeltares: I will change this.

dfastmi/kernel/core.py
MRVermeulenDeltares: I will add this, I will use capitals, since constants should be capital in python I think.

ralph-peelen-deltares:
As discussed, the sonar cloud issues are not related to the resolution

PR #24, opened 2024-02-08
DFAST-97: Resolve codesmells in kernel.py

MRVermeulenDeltares:


hrajagers:
That looks all fine to me. Sorry for the delayed review.

hrajagers (approved):
As mentioned before ... this looks fine to me.

PR #34, opened 2024-02-16
DFAST-102 improve code coverage of the batch

MRVermeulenDeltares:


tests/test_batch.py
```diff
@@ -217,5 +219,335 @@ def given_configuration_file_version_different_as_river_file_version_when_runnin
             assert str(cm.value) == 'Version number of configuration file (2.0) must match version number of rivers file (1.0)'
         finally:
             os.chdir(cwd)
+            
+class Test_batch_save_configuration_file():
```
ralph-peelen-deltares, line 223: DataAccessTest, geen GWT?

tests/test_batch.py
```diff
@@ -217,5 +219,335 @@ def given_configuration_file_version_different_as_river_file_version_when_runnin
...
+        def config(self):
+            return  ConfigParser()
+                    
+        def given_version_1_when_check_configuration_then_return_false(self, rivers : RiversObject, config : ConfigParser):
+            assert dfastmi.batch.check_configuration(rivers, config) is False
+            
+        def given_general_section_when_check_configuration_then_return_false(self, rivers : RiversObject, config : ConfigParser):
+            self.set_valid_general_section(config, "1.0")
+            
+            assert dfastmi.batch.check_configuration(rivers, config) is False
```
ralph-peelen-deltares, line 342: use assert not [function_that_returns_bool]

ralph-peelen-deltares: is nicer than 
assert [function_that_returns_bool] is False

tests/test_batch.py
```diff
@@ -217,5 +219,335 @@ def given_configuration_file_version_different_as_river_file_version_when_runnin
...
+            self.set_valid_general_section_with_q_values(config, "1.0")
+            self.add_q_section(config)
+            config.set("General", "mode", "test")
+            
+            assert dfastmi.batch.check_configuration(rivers, config) is True
+            
+        def set_valid_general_section(self, config : ConfigParser, version : str):
+            config.add_section("General")
+            config.set("General", "Version", version)
+            config.set("General", "Branch", "Bovenrijn & Waal")
```
ralph-peelen-deltares, line 371: do not use real key value, use fake if you make test value your self. I believe using real values should only be used in acceptance test to users. just make up your own name. Use team members as example, or you favorite kids tv show

tests/test_batch.py
```diff
@@ -217,5 +219,335 @@ def given_configuration_file_version_different_as_river_file_version_when_runnin
...
+            config.set("Q1", "Discharge", "1300.0")
+            config.add_section("Q2")
+            config.set("Q2", "Discharge", "1300.0")
+            config.add_section("Q3")
+            config.set("Q3", "Discharge", "1300.0")
+    
+    class Test_check_configuration_v2():
+        @pytest.fixture
+        def rivers(self):
+            return  RiversObject("dfastmi/Dutch_rivers_v2.ini")
```
ralph-peelen-deltares, line 390: I have the suspicion the text "dfastmi/Dutch_rivers_v2.ini" or "dfastmi/Dutch_rivers_v1.ini" will be used a lot. can we make a general const or something for all tests?

tests/test_batch.py
```diff
@@ -217,5 +219,335 @@ def given_configuration_file_version_different_as_river_file_version_when_runnin
...
+        }
+        return config
+    
+class Test_batch_countq():       
+    @pytest.mark.parametrize("vector_data, expected_non_empty_discharges_count", [
+        ([0.123, None, 0.456, 0.789, None], 3),
+        ([0.123, 0.123, 0.456, 0.789, 0.123], 5),
+        ([None, None, None, None, None], 0),
+    ])    
+    def given_vector_with_discharges_when_countq_then_return_amount_of_non_empty_discharges(self, vector_data: Vector, expected_non_empty_discharges_count: int):
```
ralph-peelen-deltares, line 268: ok, name...:
given_vector_with_discharges_when_countq_then_return_amount_of_non_empty_discharges

given : vector_with_discharges
when : countq
then : return_amount_of_non_empty_discharges

isn't it an improvement / more explanatory with the parameterized arguments in the test function to state the expectation in then part:

given_vector_with_discharges_when_countq_then_return_**expected**_amount_of_non_empty_discharges

given : vector_with_discharges
when : countq
then : return_**expected**_amount_of_non_empty_discharges

tests/test_batch.py
```diff
@@ -217,5 +219,335 @@ def given_configuration_file_version_different_as_river_file_version_when_runnin
...
+        1.202122
+    ])       
+    def given_input_data_with_varying_slength_when_write_report_then_expect_messages_written_in_report(self, slength :float):
+        report = StringIO()
+        reach = "reach"
+        q_location = "location"
+        q_threshold = 0.123
+        q_bankfull = 0.456
+        q_stagnant = 0.789
+        tstag = 0.101112
```
ralph-peelen-deltares, line 283: why 6 numbers behind comma? double is double right :-) 
it is test data. may be using 0.1 for first, 0.2 for second, etc sufficient? just for readability

tests/test_batch.py
```diff
@@ -217,5 +219,335 @@ def given_configuration_file_version_different_as_river_file_version_when_runnin
...
+
+        report_lines = report.getvalue().split('\n')
+                
+        prefix = 'No message found for '
+        assert prefix + 'reach' in report_lines
+        assert prefix + 'report_qthreshold' in report_lines
+        assert prefix + 'report_qbankfull' in report_lines
+        assert prefix + 'closed_barriers' in report_lines
+        assert prefix + 'char_discharge' in report_lines
+        assert report_lines.count(prefix + 'char_discharge') is 3
```
ralph-peelen-deltares, line 300: why did you use the word **is** i thought for datatypes (int, etc) we use == (or != for invert) and for types **is** like:

assert Marlon is Human
assert Marlon is AltenEmployee
assert Marlon is not Stupid

assert Marlon.Age == 27
assert Marlon.Salary == 4000
assert Marlon.HourlyRate == 87.50

assert Marlon.IsFunToWorkWith

ralph-peelen-deltares (changes requested):

tests/test_batch.py
```diff
@@ -217,5 +219,335 @@ def given_configuration_file_version_different_as_river_file_version_when_runnin
...
+        def given_version_1_when_check_configuration_then_return_false(self, rivers : RiversObject, config : ConfigParser):
+            assert dfastmi.batch.check_configuration(rivers, config) is False
+            
+        def given_general_section_when_check_configuration_then_return_false(self, rivers : RiversObject, config : ConfigParser):
+            self.set_valid_general_section(config, "1.0")
+            
+            assert dfastmi.batch.check_configuration(rivers, config) is False
+            
+        def given_general_section_with_qthreshold_when_check_configuration_then_return_false(self, rivers : RiversObject, config : ConfigParser):
+            self.set_valid_general_section(config, "1.0")
```
ralph-peelen-deltares, line 345: could self.set_valid_general_section(config, "1.0") not be in test fixture setup? why the "1.0" it is all "1.0"

tests/test_batch.py
```diff
@@ -217,5 +219,335 @@ def given_configuration_file_version_different_as_river_file_version_when_runnin
...
+        
+        @pytest.fixture
+        def config(self):
+            return  ConfigParser()
+        
+        def given_version_2_when_check_configuration_then_return_false(self, rivers : RiversObject, config : ConfigParser):
+            assert dfastmi.batch.check_configuration(rivers, config) is False
+        
+        def given_general_section_when_check_configuration_then_return_false(self, rivers : RiversObject, config : ConfigParser):
+            self.set_valid_general_section(config, "2.0")
```
ralph-peelen-deltares, line 400: self.set_valid_general_section(config, "2.0") in test fixture setup?

tests/test_batch.py
```diff
@@ -217,5 +219,335 @@ def given_configuration_file_version_different_as_river_file_version_when_runnin
...
+            return reach
+
+        def given_auto_time_true_when_get_levels_v2_then_return_expected_values(self, reach : ReachAdvanced):
+            reach.qstagnant = 4.5
+            q_threshold = 1.2
+            nwidth = 3.4
+
+            Q, apply_q, time_mi, tstag, T, rsigma, celerity = dfastmi.batch.get_levels_v2(reach, q_threshold, nwidth)
+
+            assert Q == [6.7, 8.9, 10.1]
```
ralph-peelen-deltares, line 480: should you not also check on length of Q, time, T, etc... to be same as input length of hydro_q

tests/test_batch.py
MRVermeulenDeltares: Because this is a data acces test and very much dependent on the Dutch_rivers_v1 which has this data in it

MRVermeulenDeltares: I can make my own file of  Dutch_rivers_v1.ini, but I do not think that is worth the effort

tests/test_batch.py
MRVermeulenDeltares: like a string containing the data? or just a const with the name?

tests/test_batch.py
MRVermeulenDeltares: I wasnt sure what would happen with each value during testing, so I wanted to give every variable a distinct value.
Such I started with 123, 456 and ended with the last on 101112, but yeah, they can be changed to move easily read numbers behind the comma, will update.

tests/test_batch.py
MRVermeulenDeltares: I do not know why I used is, I will update it.

unfortunately some of the above described asserts are failing :sob:

tests/test_batch.py
MRVermeulenDeltares: will update

tests/test_batch.py
MRVermeulenDeltares: It is implicitly tested here, I will add a testcase for that.

tests/test_batch.py
MRVermeulenDeltares: assert not [function_that_returns_bool]
or
assert [function_that_returns_bool] is not True

MRVermeulenDeltares: discussed, use assert not [function_that_returns_bool] for false and use assert [function_that_returns_bool] for true

tests/test_batch.py
MRVermeulenDeltares: done

tests/test_batch.py
MRVermeulenDeltares: done

tests/test_batch.py
MRVermeulenDeltares: done

tests/test_batch.py
MRVermeulenDeltares: done

tests/test_batch.py
MRVermeulenDeltares: done

tests/test_batch.py
MRVermeulenDeltares: done

tests/test_batch.py
MRVermeulenDeltares: done

ralph-peelen-deltares (approved):

PR #38, opened 2024-02-22
DFAST-121 refactor file name retriever.py and add update related tests

MRVermeulenDeltares:


dfastmi/batch/FileNameRetriever.py
```diff
@@ -29,119 +29,195 @@
...
+            The variable containing the configuration (may be None for imode = 0).
+
+        Returns
+        -------
+        filenames : Dict[Any, Tuple[str,str]]
+            Returns and empty Dict.
+        """
+        return {}
+        
+class FileNameRetrieverV1(IFileNameRetriever):
```
ralph-peelen-deltares, line 135: Is V1 the legacy class and V2 the current implementation? I suggest to use FileNameRetrieverLegacy and FileNameRetriever

dfastmi/batch/FileNameRetriever.py
```diff
@@ -29,119 +29,195 @@
 
 from typing import Optional, Union, Dict, Any, Tuple
 import configparser
+from abc import ABC, abstractmethod
+from packaging import version
 
-def get_filenames_version1(
-    config: Optional[configparser.ConfigParser] = None,
-) -> Dict[Any, Tuple[str,str]]:
+class IFileNameRetriever(ABC):
```
ralph-peelen-deltares, line 35: Why IFileNameRetriever and not AFileNameRetriever, there is a protected method in this class..

dfastmi/batch/FileNameRetriever.py
```diff
@@ -29,119 +29,195 @@
...
+
+        Returns
+        -------
+        value : str
+            The value specified for the key in the chapter.
+        """
+        try:
+            return config[chap][key]
+        except:
+            raise Exception('Keyword "{}" is not specified in group "{}" of analysis configuration file.'.format(key,chap,))
```
ralph-peelen-deltares, line 84: please use (for me more readable) fstring:

```suggestion
            raise Exception(f'Keyword "{key}" is not specified in group "{chap}" of analysis configuration file.')
```

dfastmi/batch/FileNameRetriever.py
```diff
@@ -29,119 +29,195 @@
...
+
+def get_filename_retriever(imode : int, config : configparser.ConfigParser, need_tides : bool) -> IFileNameRetriever:
```
ralph-peelen-deltares, line 86: this is the factory right? Maybe create factory class and use a register function with dictionary and creator and a retrieve function to get correct creator or a default: 
(this is a out of my head example
```
class FileNameRetrieverFactory:
    """
    Class is used to register and get creation of AFileNameRetriever Objects
    """
    _creators = {}
    """Contains the AFileNameRetriever Objects creators to be used"""

    def __init__(self):
        self._creators = {}

    def register_creator(self, imode: int, creator: Callable[bool, AFileNameRetriever]):
        """Register creator function to create a AFileNameRetriever object."""
        self._creators[imode] = creator

    def generate(self, imode: int, needs_tide: bool) -> AFileNameRetriever:
        """Call the Constructor function to generate AFileNameRetriever object."""
        constructor = self._constructors.get(imode)
        if constructor:
            return constructor(needs_tide)
        else:
            return FileNameRetrieverUnsupported()

Comment (ralph-peelen-deltares) [reply dfastmi/batch/FileNameRetriever.py:86]: and in the call class register mode when read / created (constructor / initializer core.py ?)
FileNameRetrieverFactory.register_creator(1, lambda needs_tide: FileNameRetrieverLegacy() )
FileNameRetrieverFactory.register_creator(2, lambda needs_tide: FileNameRetriever(needs_tide))

File: dfastmi/batch/FileNameRetriever.py [hunk 5]
```diff
@@ -29,119 +29,195 @@
...
+        filenames : Dict[Any, Tuple[str,str]]
+            Dictionary of string tuples representing the D-Flow FM file names for
+            each reference/with measure pair. The keys of the dictionary vary. They
+            can be the discharge index, discharge value or a tuple of forcing
+            conditions, such as a Discharge and Tide forcing tuple.
+        """
+        filenames: Dict[Any, Tuple[str,str]]
+        filenames = {}
+        for i in range(3):
+            qstr = "Q{}".format(i + 1)
```
ralph-peelen-deltares, line 161: use fstring?
```suggestion
            qstr = f"Q{i+1}"
```

dfastmi/batch/FileNameRetriever.py
```diff
@@ -29,119 +29,195 @@
...
+            can be the discharge index, discharge value or a tuple of forcing
+            conditions, such as a Discharge and Tide forcing tuple.
+        """
+        filenames: Dict[Any, Tuple[str,str]]
+        key: Union[Tuple[float, int], float]
+        filenames = {}
+        i = 0
+        while True:
+            i = i + 1
+            CSTR = "C{}".format(i)
```
ralph-peelen-deltares, line 209: use fstring? 
```suggestion
            CSTR = f'C{i + 1}'
```

dfastmi/batch/FileNameRetriever.py
```diff
@@ -29,119 +29,195 @@
...
+            if CSTR in config:
+                Q = float(self._cfg_get(config, CSTR, "Discharge"))
```
ralph-peelen-deltares, line 211: niet except erbij oid... float parsing van string ....

ralph-peelen-deltares (changes requested):
see comments

dfastmi/batch/FileNameRetriever.py
MRVermeulenDeltares: todo

dfastmi/batch/FileNameRetriever.py
MRVermeulenDeltares: todo

dfastmi/batch/FileNameRetriever.py
MRVermeulenDeltares: how do we do protected methods?

MRVermeulenDeltares: done, question in regard to protected method still stands.

dfastmi/batch/FileNameRetriever.py
MRVermeulenDeltares: done

dfastmi/batch/FileNameRetriever.py
MRVermeulenDeltares: done

dfastmi/batch/FileNameRetriever.py
```diff
@@ -27,121 +27,210 @@
 This file is part of D-FAST Morphological Impact: https://github.com/Deltares/D-FAST_Morphological_Impact
 """
```
ralph-peelen-deltares, line 28: Because this module contains multiple classes would the following comment / header info handy?
```suggestion
"""
"""
Module for Abstract class AFileNameRetriever, FileNameRetriever implementations and the FileNameRetrieverFactory classes

Abstract Class:
    AFileNameRetriever
    
Class: 
    FileNameRetrieverUnkown
    FileNameRetrieverLegacy
    FileNameRetriever

Static Class: 
    FileNameRetrieverFactory

"""
```

tests/batch/test_FileNameRetriever.py
```diff
@@ -0,0 +1,195 @@
...
+class Test_FileNameRetriever():
+    class Test_FileNameRetrieverFactory():
+        @pytest.fixture
+        def factory(self) -> FileNameRetriever.FileNameRetrieverFactory:
+            factory = FileNameRetriever.FileNameRetrieverFactory()
+            factory.register_creator("1.0", lambda needs_tide: FileNameRetriever.FileNameRetrieverLegacy())
+            factory.register_creator("2.0", lambda needs_tide: FileNameRetriever.FileNameRetriever(needs_tide))
+            return factory
+        
+        def given_version_1_with_varying_needs_tide_when_generate_then_return_FileNameRetrieverLegacy(self, factory : FileNameRetriever.FileNameRetrieverFactory):
```
ralph-peelen-deltares, line 14: Unclear because I forgot.. but wasn't it agreed between us to add a text here splitting the name in GWT:
```suggestion
     """"
     given : version 1 with varying needs tide
     when :  generate 
     then  : FileNameRetrieverLegacy
     """
```

dfastmi/batch/FileNameRetriever.py
```diff
@@ -27,121 +27,210 @@
...
-            else:
-               key = Q
-            filenames[key] = (reference, measure)
+    _creators = {}
+    """Contains the AFileNameRetriever Objects creators to be used"""
+
+    def __init__(self):
+        self._creators = {}
+
+    def register_creator(self, version: str, creator: Callable[[bool], AFileNameRetriever]):
```
ralph-peelen-deltares, line 95: May be use Version object instead of a str... 
str of "2.0" is not same as "2"... but i think version("2.0 ") == version("2")

tests/batch/test_FileNameRetriever.py
```diff
@@ -0,0 +1,195 @@
...
+        @pytest.mark.parametrize("use_tide", [
+            True,
+            False
+        ])
+        def given_partial_setup_config_parser_when_get_file_names_then_throw_exception_with_expected_message(self, use_tide):
+            file_name_retriever = FileNameRetriever.FileNameRetriever(use_tide)
+            
+            key = "Discharge"
+            chap = "C1"
+            expected_exception_message = 'Keyword "{}" is not specified in group "{}" of analysis configuration file.'.format(key, chap)
```
ralph-peelen-deltares, line 107: please use f string
```suggestion
            expected_exception_message = f'Keyword "{key}" is not specified in group "{chap}" of analysis configuration file.'
```

ralph-peelen-deltares (changes requested):
Paar kleine dingetjes

dfastmi/batch/FileNameRetriever.py
MRVermeulenDeltares: agreed

dfastmi/batch/FileNameRetriever.py
MRVermeulenDeltares: I tried to get it working before, but since the retrieved version from the config is a string it didnt work, will try it again by making them both version and see if I can get that to work.

MRVermeulenDeltares: done

dfastmi/batch/FileNameRetriever.py
```diff
@@ -27,121 +27,227 @@
 This file is part of D-FAST Morphological Impact: https://github.com/Deltares/D-FAST_Morphological_Impact
 """
 
-from typing import Optional, Union, Dict, Any, Tuple
+"""
```
MRVermeulenDeltares, line 30: discussed, move classes into seperate files.

ralph-peelen-deltares (approved):

PR #45, opened 2024-03-04
DFAST-106: Update user manual

hrajagers:
Update user manual
Update Dutch_rivers_v2.ini to better match the intended final version


hrajagers:
@MRVermeulenDeltares, @ralph-peelen-deltares I'm not sure why this branch fails the code analysis since (in the end) there are no code adjustments. Maybe it's because some commit touched the code (I did a merge of all changes in the master).

ralph-peelen-deltares:
@hrajagers don't be scared. we discussed this together, the problem is the merge main commit. This should and will be changed to update branch with rebase main. This will not happen again :-)

PR #47, opened 2024-03-05
DFAST-94 Refactor char_discharge and char_times in kernel into a new class (legacy)

MRVermeulenDeltares:


ralph-peelen-deltares (approved):
Move done correctly. New file is legacy, will not refactor this further for now. Not in scope

PR #50, opened 2024-03-12
DFAST-110 remove usage of FileUtils and related tests

ralph-peelen-deltares:


dfastmi/batch/DFastUtils.py
```diff
@@ -132,3 +133,14 @@ def _get_km_bins(km_bin: Tuple[float, float, float], type_characteristic_chainag
     km = km_bin[0] + dx + numpy.arange(lb, ub) * km_step
 
     return km
+
+def get_progloc() -> str:
```
priscavdsluis, line 137: Please use google doc strings as agreed

priscavdsluis, line 137: Perhaps return a `Path` object instead of string as we want to work with Paths instead of strings.

MRVermeulenDeltares: is that something we can tackle codewide with black?

dfastmi/batch/DFastUtils.py
```diff
@@ -132,3 +133,14 @@ def _get_km_bins(km_bin: Tuple[float, float, float], type_characteristic_chainag
...
+    """
+    Get the location of the program.
+
+    Arguments
+    ---------
+    None
+    """
+    progloc = str(pathlib.Path(__file__).parent.parent.absolute())
```
priscavdsluis, line 145:
```suggestion
    import dfastmi
    progloc = str(pathlib.Path(dfastmi.__file__).parent.absolute())
```

This is more reliable

tests/io/test_DFastUtils.py
```diff
@@ -0,0 +1,10 @@
+
+from dfastmi.batch.DFastUtils import get_progloc
+
+class Test_DFastUtils():
+    class Test_get_progloc():
+        def test_get_program_location(self):
+            """
+            Get the location of the program.
+            """
+            assert get_progloc() != ""
```
priscavdsluis, line 10:
```suggestion

from dfastmi.batch.DFastUtils import get_progloc

class TestGetProgloc():
    def test_get_program_location(self):
        """
        Get the location of the program.
        """
        assert get_progloc().endswith("dfastmi")
```

dfastmi/batch/DFastUtils.py
priscavdsluis: After discussion with @MRVermeulenDeltares I think it might be best to adhere to the currently used style. As far I have seen up until now the most used style is Numpydoc:
https://numpydoc.readthedocs.io/en/latest/format.html

So lets be consistent with that.

ralph-peelen-deltares: Will update but can you show me what I did wrong? I moved this code...

ralph-peelen-deltares: Yes good idea. Will update

dfastmi/batch/DFastUtils.py
ralph-peelen-deltares: Wow, didn't know this. Will use the last one, this seems the most stable

tests/io/test_DFastUtils.py
ralph-peelen-deltares: Nice, will update. Thanks for the commit possibility

PR #51, opened 2024-03-12
DFAST-130: Validation in Reach class

ralph-peelen-deltares:


dfastmi/io/CelerObject.py
```diff
@@ -37,36 +37,52 @@
...
-from abc import ABC, abstractmethod
-from typing import List
+from abc import ABC
+from typing import List, Tuple
+from pydantic import BaseModel
+from dfastmi.io.AReach import AReach
 
-
-class ICelerObject(ABC):
+class ICelerObject(ABC, BaseModel):
```
priscavdsluis, line 45: Interfaces in Python aren't really a thing. Python allows for multi inheritance, and you should use abstract or base classes for this.
It might sound like potato potato, but I think it is important to adhere to a certain standard.

priscavdsluis, line 45: I suggest introducing a `DFastBaseModel` that inherits from pydantic's BaseModel to configure this class for all of D-FAST, so that all of our objects such as CelerObject, Branch, etc. can inherit from this and all are configured similarly.

pyproject.toml
```diff
@@ -21,6 +21,7 @@ mkl = "2021.4.0"
 geopandas = "^0.14.2"
 matplotlib = "^3.6.3"
 shapely = "^2.0.2"
+pydantic = "^1.9.0"
```
priscavdsluis, line 24: Why not use the latest pydantic? The latest version is already 2.6 :)

dfastmi/io/CelerObject.py
```diff
@@ -37,36 +37,52 @@
...
-            #                 reach,
-            #             )
-            #         )
-            return
+    """
+        Series of discharge, bed celerity pairs
+    """
+    cdisch : Tuple[float,float] = (0.0, 0.0)
     
+    def validate(self, values):
```
priscavdsluis, line 55: if you want to use automatic validation from Pydantic, you should decorate your functions with the `@validator` (v1) or `@model_validator` (v2) decorator.

V2 https://docs.pydantic.dev/latest/concepts/validators/#model-validators
V1 https://docs.pydantic.dev/1.10/usage/validators/

dfastmi/io/CelerObject.py
```diff
@@ -37,36 +37,52 @@
...
+        branch_name : str = self.parent_reach.parent_branch.name
+        reach_name : str = self.parent_reach.name
+        if self.cdisch == (0.0, 0.0):
+             raise ValueError(f'The parameter "CelerQ" must be specified for branch "{branch_name}", '
+                              f'reach "{reach_name}" since "CelerForm" is set to 2.')
```
priscavdsluis, line 60: The information in this message does not correspond with the check that you're doing?

dfastmi/io/CelerObject.py
priscavdsluis: With V2 you could write this as:

```
    @model_validator(mode='after')
    def validate(self) -> 'CelerDischarge':
        branch_name : str = self.parent_reach.parent_branch.name
        reach_name : str = self.parent_reach.name
        if self.cdisch == (0.0, 0.0):
             raise ValueError(f'The parameter "CelerQ" must be specified for branch "{branch_name}", '
                              f'reach "{reach_name}" since "CelerForm" is set to 2.')
        
        return self
```

The validation is automatically called directly after the initialization of the `CelerDischarge` object, and call to the super class is not needed, since validators are inherited by default.

pyproject.toml
ralph-peelen-deltares: will update

dfastmi/io/CelerObject.py
ralph-peelen-deltares: Thanks! really great... I thought i was doing it wrong... will update

dfastmi/io/CelerObject.py
ralph-peelen-deltares: This is the original warning. I believe it is correct. The default is (0.0, 0.0) if this is needed in the configuration because CelerForm is set to 2 and not available in the river configuration or available but set to this default we need to report/feedback this message.

dfastmi/io/CelerObject.py
ralph-peelen-deltares: Lets have a team meeting/discussion about this and resolve for multiple objects in new issue.

ralph-peelen-deltares: I will try and do this. Good suggestion!

PR #52, opened 2024-03-13
DFAST-147: Update package name in pyproject.toml to dfastmi

priscavdsluis:


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [tool.poetry]
-name = "d-fast_morphological_impact"
+name = "dfastmi"
 version = "2.0.8"
```
MRVermeulenDeltares, line 3: should this also be ""3.0.0"

Still little bit unclear to me why this is incorrect.

priscavdsluis: Not sure, I can update the version, Not sure how our versioning works :)

MRVermeulenDeltares: I am saying it because @ralph-peelen-deltares stated it in the example of discord, not sure how this would directly impact expected versioning in the software...

MRVermeulenDeltares: discussed, working for Prisca.

MRVermeulenDeltares (approved):

PR #62, opened 2024-03-26
DFAST-142: Update year to 2024 in copyright

priscavdsluis:


ralph-peelen-deltares (approved):

MRVermeulenDeltares (approved):
Looks good, question regarding the (C), is this the correct font in regard to the actual copyright icon? ©

PR #64, opened 2024-03-28
DFAST-142: Revert copyright symbol change

priscavdsluis:


MRVermeulenDeltares (approved):
Teamcity seems to work again, Nuitka doesnt appear to work with ©
We might want to look into it in the future to resolve this in some way, if it is visible somewhere in the application or settings.

PR #69, opened 2024-04-03
DFAST-138: suppress the command window

ralph-peelen-deltares:


MRVermeulenDeltares (approved):
Few Questions before we should merge this:
1. We now will have to .bat files to build, we tried to prevent this at the start of the project, this is in need of a settings or make file of the like. How and when do we tackle this?
2. What are we gonna do with CLI, we cant really distribute this suppressed cmd windows .exe if we have no clear vision and decision on this.
3. How should this issue be tested?
4. What needs to be updated on teamcity?
if this will become the distributed version, we will need to couple it to the signed release, would that be all changes on teamcity to make it a viable release?

Please try to answer these questions before you merge it to main.

ralph-peelen-deltares:
> Few Questions before we should merge this:
> 
> 1. We now will have to .bat files to build, we tried to prevent this at the start of the project, this is in need of a settings or make file of the like. How and when do we tackle this?
> 2. What are we gonna do with CLI, we cant really distribute this suppressed cmd windows .exe if we have no clear vision and decision on this.
> 3. How should this issue be tested?
> 4. What needs to be updated on teamcity?
>    if this will become the distributed version, we will need to couple it to the signed release, would that be all changes on teamcity to make it a viable release?
> 
> Please try to answer these questions before you merge it to main.
1) We need to do this during refinement where we discuss this and create issues. PO in combination with PL can prioritize the request. Another moment could be by stating it in our weekly progress meeting with the PO/PL before refining. I am not aware of any agreements that we should prevent making any bat files. When was this decided?
2) As discussed in weekly progress meeting we make both in the scope of this issue. Later PO/PL will decide what to do with CLI
3) Download signed build and during start of dfastmi.exe you will always see a command prompt:
image. This should be gone.
4) Yes indeed we also need a copy of the signed release. Will create and update. As discussed during weekly progress meeting and apparently poorly written down in the comments of the issue, currently in scope is both releases.

PR #72, opened 2024-04-03
DFAST-124 refactor detect and plot.py

MRVermeulenDeltares:


dfastmi/batch/AreaDetector.py
```diff
@@ -0,0 +1,389 @@
...
+
+    @property
+    def area_list(self) -> list:
+        """
+        List of sub areas
+        """
+        return self._area_list
+
+    @property
+    def total_area_weigth(self) -> numpy.ndarray:
```
tim-vd-aardweg, line 59: typo: weigth -> weight

dfastmi/batch/AreaDetector.py
```diff
@@ -0,0 +1,389 @@
...
+        edgeface_indeces: numpy.ndarray,
+        areai: numpy.ndarray,
+        wbin: numpy.ndarray,
+        wthresh: numpy.ndarray,
+        siface: numpy.ndarray,
+        afrac: numpy.ndarray,
+        sbin: numpy.ndarray,
+        sthresh: numpy.ndarray,
+        slength: float,
+    ):
```
tim-vd-aardweg, line 84: Missing return typehint

dfastmi/batch/AreaDetector.py
```diff
@@ -0,0 +1,389 @@
...
+INFORMATION
+This file is part of D-FAST Morphological Impact: https://github.com/Deltares/D-FAST_Morphological_Impact
+"""
+import math
+from typing import Tuple
+
+import numpy
+
+
+class AreaDetector:
```
tim-vd-aardweg, line 35: Maybe add some class documentation explaining what this class does?

dfastmi/batch/AreaDetector.py
```diff
@@ -0,0 +1,389 @@
...
+            volume[2, ia], area[ia], volume[0, ia] = self._comp_sedimentation_volume2(
+                numpy.maximum(dzgemi_filtered, 0.0), dzmin, areai, slength, nwidth
+            )
+
+        sorted_list = numpy.argsort(area)[::-1]
+        area = area[sorted_list]
+        volume = volume[:, sorted_list]
+        sub_area_list = [sub_area_list[ia] for ia in sorted_list]
+
+        self._area = area
```
tim-vd-aardweg, line 126: I think I'd prefer to use a DTO here. Something like `DetectedAreas` that has three properties. The `detect_areas` can return this DTO. It may also conceptually make more sense. For example: Instead of passing a AreaDetector to the `ErosionAreaPlotter`, you can only pass the data, the `DetectedArea`.

dfastmi/batch/AreaDetector.py
```diff
@@ -0,0 +1,389 @@
...
+        return self._total_area_weigth
+
+    def __init__(self):
+        self._area: numpy.ndarray = numpy.zeros(0)
+        self._volume: numpy.ndarray = numpy.zeros(0)
+        self._area_list: list = []
+        self._total_area_weigth: numpy.ndarray = numpy.zeros(0)
+
+    def detect_areas(
+        self,
```
tim-vd-aardweg, line 72: Missing documentation

dfastmi/batch/AreaDetector.py
```diff
@@ -0,0 +1,389 @@
...
+        areai: numpy.ndarray,
+        wbin: numpy.ndarray,
+        wthresh: numpy.ndarray,
+        siface: numpy.ndarray,
+        afrac: numpy.ndarray,
+        sbin: numpy.ndarray,
+        sthresh: numpy.ndarray,
+        slength: float,
+    ):
+        self._total_area_weigth = numpy.zeros(dzgemi.shape)
```
tim-vd-aardweg, line 85: typo in weight

dfastmi/batch/AreaDetector.py
```diff
@@ -0,0 +1,389 @@
...
+
+        sbin_length = sthresh[1] - sthresh[0]
+        nwidth = wthresh[-1] - wthresh[0]
+        sub_areai, n_sub_areas = self._detect_connected_regions(
+            dzgemi > dzmin, edgeface_indeces
```
tim-vd-aardweg, line 90: Isnt the plural 'indexes' or 'indices'? :D

dfastmi/batch/AreaDetector.py
```diff
@@ -0,0 +1,389 @@
...
+    def total_area_weigth(self) -> numpy.ndarray:
+        """
+        total area weigth
+        """
+        return self._total_area_weigth
+
+    def __init__(self):
+        self._area: numpy.ndarray = numpy.zeros(0)
+        self._volume: numpy.ndarray = numpy.zeros(0)
+        self._area_list: list = []
```
tim-vd-aardweg, line 68: Any reason this is a plain list and not a numpy array?

dfastmi/batch/AreaDetector.py
```diff
@@ -0,0 +1,389 @@
...
+                dzgemi_filtered,
+                dzmin,
+                areai,
+                wbin,
+                siface,
+                afrac,
+                sbin,
+                wthresh,
+                slength,
+                sbin_length,
```
tim-vd-aardweg, line 113: `sbin_length` is a `ndarray[Any, dtype]` while the function it is passed to is expecting a `float`. Are you sure this is correct?

dfastmi/batch/AreaDetector.py
```diff
@@ -0,0 +1,389 @@
...
+            )
+            self._total_area_weigth += wght_area_ia
+
+            volume[2, ia], area[ia], volume[0, ia] = self._comp_sedimentation_volume2(
+                numpy.maximum(dzgemi_filtered, 0.0), dzmin, areai, slength, nwidth
```
tim-vd-aardweg, line 118: Is `nwidth` a float?

dfastmi/batch/AreaDetector.py
MRVermeulenDeltares: The area is based on sub_area_list, sub_area_list is also initialized as a empty list.

dfastmi/batch/AreaDetector.py
MRVermeulenDeltares: I see nwidth = wthresh[-1] - wthresh[0],  since wtrash is an ndarray, nwidth could basicly be anything 💀

dfastmi/batch/AreaPlotter.py
```diff
@@ -0,0 +1,388 @@
...
+        self,
+        plotting_options: PlotOptions,
+        plot_n: int,
+        area_detector: AreaDetector,
+        xyz_file_location: Path,
+    ):
+        """
+        Arguments
+        ---------
+        plotting_options : PlotOptions
```
tim-vd-aardweg, line 63: Arguments are in the wrong order in the docs :D

dfastmi/batch/AreaDetector.py
MRVermeulenDeltares: assuming nwidth stands for normal_width, which is also know in reach as a float I assume it should be float.
Based on the information in this method it could be anything.

dfastmi/batch/AreaPlotter.py
```diff
@@ -0,0 +1,388 @@
...
+            areai,
+            wbin,
+            siface,
+            afrac,
+            sbin,
+            wthresh,
+            sthresh,
+        )
+
+        self._write_xyz_file(wbin_labels, kmid, binvol)
```
tim-vd-aardweg, line 129: Typehinting for `wbin_labels` may be incorrect. `list[str]` vs `ndarray[Any, dtype]`

tim-vd-aardweg, line 129: This is a sneaky little side-effect of the `plot_areas` function. Maybe we can check if we can move this out of this function and place at a higher level.

MRVermeulenDeltares: agreed

dfastmi/batch/AreaDetector.py
MRVermeulenDeltares: fixed

dfastmi/batch/AreaDetector.py
MRVermeulenDeltares: checked together, now tpehinted as float.

dfastmi/batch/AreaDetector.py
MRVermeulenDeltares: lets discuss

dfastmi/batch/AreaPlotter.py
```diff
@@ -0,0 +1,388 @@
...
+            siface,
+            afrac,
+            sbin,
+            sthresh,
+            kmid,
+            sbin_length,
+            binvol,
+        )
+
+    def _write_xyz_file(
```
tim-vd-aardweg, line 145: Maybe make a separate XYZ writer?

dfastmi/batch/AreaPlotter.py
MRVermeulenDeltares: I checked, it seems list[str] lets check together.

dfastmi/batch/AreaPlotter.py
```diff
@@ -0,0 +1,388 @@
...
+            areai,
+            wbin,
+            wbin_labels,
+            wthresh,
+            siface,
+            afrac,
+            sbin,
+            sthresh,
+            kmid,
+            sbin_length,
```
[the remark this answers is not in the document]
MRVermeulenDeltares: fix to typehint :float

dfastmi/batch/AreaPlotter.py
MRVermeulenDeltares: fixed

dfastmi/batch/AreaPlotter.py
tim-vd-aardweg, line 1: We shortly discussed the responsibilities of this class. We agreed it does too much, but it is already so much better than it used to be. Further refactoring/restructuring may be done in follow-up issues if required.

dfastmi/batch/AreaPlotter.py
```diff
@@ -0,0 +1,388 @@
...
+    def _save_figure(self, fig: Figure, ax: Axes, figure_base_name: str):
+        if not self._plotting_options.saveplot:
+            return
+
+        figbase = self._plotting_options.figure_save_directory / figure_base_name
+        if self._plotting_options.saveplot_zoomed:
+            dfastmi.plotting.zoom_x_and_save(
+                fig,
+                ax,
+                figbase,
```
tim-vd-aardweg, line 288: You are passing a path, while a string is expected.

dfastmi/batch/AreaDetector.py
MRVermeulenDeltares: done

dfastmi/batch/AreaDetector.py
MRVermeulenDeltares: done

dfastmi/batch/AreaDetector.py
MRVermeulenDeltares: done

dfastmi/batch/AreaDetector.py
MRVermeulenDeltares: now AreaData

dfastmi/batch/AreaDetector.py
MRVermeulenDeltares: now AreaData

dfastmi/batch/AreaDetector.py
MRVermeulenDeltares: now initialized as area_list: list[bool] = field(default_factory=list)

tests/batch/test_AreaPlotter.py
```diff
@@ -0,0 +1,351 @@
...
+        area_detector = mock_area_detector()
+        plot_n = 3
+
+        plotter = SedimentationAreaPlotter(plotting_options, plot_n, area_detector)
+
+        assert plotter._area_str == "sedimentation area {}"
+        assert plotter._total_str == "total sedimentation volume"
+        assert plotter._positive_up
+
+    def test_plot_areas_with_plotting_plots_areas(self, setup, tmp_path: Path):
```
tim-vd-aardweg, line 88: What is it you are actually testing here? You are not testing if the plots are actually being plotted, only if some of the implementation details are working as expected. For me as a complete DFAST noob, it means nothing that `plot_sedimentation` is being called 10 times. Why even 10 times? What if we change the implementation details, would we update this number to 9 or 11?

Wouldn't we get way more value out of a single integration / data access test that actually tests if the images are being created? That only breaks if/when the images are not being created?

dfastmi/batch/AreaPlotter.py
MRVermeulenDeltares: done

dfastmi/batch/AreaPlotter.py
MRVermeulenDeltares: as discussed, moved the xyz writer out of it.

dfastmi/batch/AreaPlotter.py
MRVermeulenDeltares: 1csU

MRVermeulenDeltares: gif

dfastmi/batch/AreaPlotter.py
```diff
@@ -0,0 +1,388 @@
...
+                self._plotting_options.plot_extension,
+                self._plotting_options.kmzoom,
+            )
+        figfile = figbase.with_suffix(self._plotting_options.plot_extension)
+        dfastmi.plotting.savefig(fig, figfile)
```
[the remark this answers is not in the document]
MRVermeulenDeltares: See comment above.

tests/batch/test_AreaPlotter.py
MRVermeulenDeltares: The test verifies the generation of 10 plots: one for the total volume and nine for the areas. Its focus is solely on this unit and does not necessitate accurate data for testing purposes. Instead, it should successfully relay its data to the plotter, which in turn will execute the plotting process.

The unit test should not encompass the testing of the actual plotting mechanism. Its primary objective is to validate that the plotter is invoked based on the provided input data. I can refine the unit test to be more specific in its assertions, this might make it more prone to breaking with changes in the implementation.

We could discard these unit tests and explore an integration test approach, wherein we compare the output images. To achieve this, we would need to implement image processing, which presents its own pros and cons. Alternatively, we could consider utilizing Ranorex for this purpose.

Would you like to proceed with abandoning the unit tests and exploring image processing for testing? If so, I'll remove the file from the branch. Should the image processing be implemented on this branch in that case?

dfastmi/batch/AreaPlotter.py
```diff
@@ -0,0 +1,388 @@
...
+from dfastmi.batch.AreaDetector import AreaDetector
+from dfastmi.batch.PlotOptions import PlotOptions
+
+
+class AreaPlotter(ABC):
+    """
+    Abstract class used to plot the area.
+    """
+
+    _total_str: str
```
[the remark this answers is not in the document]
tim-vd-aardweg: What about `x_axis_label` and `y_axis_label_`? Or if it is the title of the graph: `graph_title`? :D

tests/batch/test_AreaPlotter.py
tim-vd-aardweg: For now it is fine :) My main confusion was about where the 10 was comming from. Maybe change the magic number into a variable to describe what it is?

dfastmi/batch/AreaPlotter.py
```diff
@@ -0,0 +1,388 @@
...
+        sbin_length: float,
+        binvol: List[numpy.ndarray],
+    ):
+        if not self._plotting_options.plotting:
+            return
+
+        fig, ax = dfastmi.plotting.plot_sedimentation(
+            kmid,
+            "chainage [km]",
+            binvol,
```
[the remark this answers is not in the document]
MRVermeulenDeltares: binvol is a List[numpy.ndarray], in this call a numpy.ndarray is expected, no change in binvol as far as I know in this issue.

dfastmi/batch/AreaPlotter.py
MRVermeulenDeltares: Do we want to apply path on plotting or str on the call of the path methods for now?

dfastmi/batch/AreaPlotter.py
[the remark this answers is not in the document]
MRVermeulenDeltares: renamed to _graph_title_of_the_total_area and _template_graph_title_of_the_sub_areas

dfastmi/batch/AreaPlotter.py
MRVermeulenDeltares: fixed

dfastmi/batch/AreaPlotter.py
tim-vd-aardweg: My preference is to have the Path object as long as possible. However, that may be out of scope for this issue.

PR #76, opened 2024-04-05
DFAST-197: Safely retrieve keys from config

ralph-peelen-deltares:
- all string have empty string as fallback
- all bools have false as fallback

### DIFF
--- a/dfastmi/gui/dialog_model.py
+++ b/dfastmi/gui/dialog_model.py
@@ -106,7 +106,7 @@ class DialogModel:
     @property
     def case_description(self) -> str:
         """Get case description."""
-        return self.section["CaseDescription"]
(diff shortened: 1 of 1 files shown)


dfastmi/gui/dialog_model.py
```diff
@@ -106,7 +106,7 @@ def __init__(
     @property
     def case_description(self) -> str:
         """Get case description."""
-        return self.section["CaseDescription"]
+        return self.section.get("CaseDescription", "")
```
MRVermeulenDeltares, line 109: how would these empty fallbacks look like in the UI?

MRVermeulenDeltares (commented):
Do we have some test cases we want to do before merging this?
Shall we wait on the signed build to test this.

PR #82, opened 2024-04-05
DFAST-170: Write conditions and simulation files to the report file

ralph-peelen-deltares:


dfastmi/batch/core.py
```diff
@@ -190,9 +191,47 @@ def _report_analysis_configuration(
     _report_basic_analysis_configuration(
         branch, reach, q_threshold, ucrit, slength, report
     )
+    _report_discharge_branch_location(branch.qlocation, report)
+    _report_analysis_conditions_header(report)
+    for section in config:
```
tim-vd-aardweg, line 196: Make a separate method for this, like with the other parts of the log file?

MRVermeulenDeltares: Resolved.

PR #83, opened 2024-04-08
DFAST-199: When loading DFast MI configuration checkboxs are not enabled

ralph-peelen-deltares:
- not all plot controls are enabled


MRVermeulenDeltares (approved):
Looks ok, first look with debug mode seems to resolve the problem.
I found something that could be a new bug, I will create a new issue for that.

when save figures is ticked, but make figures is unticked, the save location is still writeable.
python_9g4Wkco1zS

MRVermeulenDeltares (approved):

PR #85, opened 2024-04-10
DFAST-200: model should return empty string when RiverKM key is not in general section of DFAST MI config file

ralph-peelen-deltares:


MRVermeulenDeltares (approved):
Change looks OK.
Verified in Debug mode that Qmin_4000_v2.cfg, Qmin_4000_v2_rkm.cfg and Qmin_4000_v2_tide.cfg could run.

PR #86, opened 2024-04-10
DFAST-201 When disabling make figures with save figures enabled, the figure directory stays editable in the GUI

MRVermeulenDeltares:


dfastmi/gui/dialog_view.py
```diff
@@ -1069,32 +1077,29 @@ def _update_plotting(self) -> None:
...
+        figure_dir_button.setEnabled(value)
 
     def _update_save_plotting(self) -> None:
         """Update the plotting flags."""
 
-        save_plot_gui = (
-            self._save_plots_edit.isChecked() and self._view_model.model.plotting
-        )
-
-        if self._view_model.model.save_plots != save_plot_gui:
```
ralph-peelen-deltares, line 1089: this line should remain i think.

dfastmi/gui/dialog_view.py
```diff
@@ -851,7 +857,9 @@ def _menu_load_configuration(self) -> None:
                     )
                 )
             else:
-                self._update_enabling_plotting_controls()
+                self._update_enabled_of_make_plot_dependent_view_items(
```
ralph-peelen-deltares, line 860: this should be removed. The setting of the enabled flags should be done in the view model in load_config

dfastmi/gui/dialog_view.py
MRVermeulenDeltares: reverted back with the discussed change.

dfastmi/gui/dialog_view.py
MRVermeulenDeltares: updated in pair programming

MRVermeulenDeltares:
Tested with running the following cfgs in debug mode:
Qmin_4000_v2.cfg, Qmin_4000_v2_rkm.cfg & Qmin_4000_v2_tide.cfg.

everything seems to function correctly.

PR #89, opened 2024-05-30
Code updates required for formal 3.0 release

hrajagers:
* command line help adjusted to focus only the primary functionality (don't include backward compatibility stuff)
* add names of simulation files to config file

### DIFF
--- a/dfastmi/__main__.py
+++ b/dfastmi/__main__.py
@@ -121,7 +121,7 @@ def parse_arguments() -> Tuple[str, str, Optional[str], str, bool]:
 
     parser.add_argument(
         "--reduced_output",
-        help=help=argparse.SUPPRESS,
+        help=argparse.SUPPRESS,
         action="store_true",
     )
     parser.set_defaults(reduced_output=False)
--- a/dfastmi/__main__.py
+++ b/dfastmi/__main__.py
@@ -94,30 +94,34 @@ def parse_arguments() -> Tuple[str, str, Optional[str], str, bool]:
     parser = argparse.ArgumentParser(description="D-FAST Morphological Impact.")
 
     parser.add_argument(
-        "--language", help="display language 'NL' or 'UK' (UK is default)"
+        "--language",
+        choices=["UK","NL"],
+        default="UK",
+        help=argparse.SUPPRESS,
     )
-    parser.set_defaults(language="UK")
 
     parser.add_argument(
-        "--mode", help="run mode 'BATCH', 'CLI' or 'GUI' (GUI is default)"
+        "--mode",
+        choices=["CLI","BATCH","GUI"],
+        default="GUI",
+        help="run mode 'BATCH' or 'GUI' (%(default)s is default)",
     )
-    parser.set_defaults(mode="GUI")
 
     parser.add_argument(
         "--config",
-        help="name of analysis configuration file ('dfastmi.cfg' is default)",
+        default="dfastmi.cfg",
+        help="name of analysis configuration file ('%(default)s' is default)",
     )
-    parser.set_defaults(config="dfastmi.cfg")
 
     parser.add_argument(
         "--rivers",
-        help="name of rivers configuration file ('Dutch_rivers.ini' is default)",
+        default="unspecified",
+        help="name of rivers configuration file ('Dutch_rivers_v2.ini' is default)",
     )
-    parser.set_defaults(rivers="unspecified")
 
     parser.add_argument(
         "--reduced_output",
-        help="write reduced M/N range (structured model only)",
+        help=help=argparse.SUPPRESS,
         action="store_true",
     )
     parser.set_defaults(reduced_output=False)
--- a/dfastmi/gui/dialog_model.py
+++ b/dfastmi/gui/dialog_model.py
@@ -291,18 +291,17 @@ class DialogModel:
         measure_files: List,
     ) -> None:
         """Get condition configuration."""
-        num_files = min(len(reference_files), len(measure_files))
-
-        for i, discharge in enumerate(reach.hydro_q[:num_files]):
-            if discharge in reference_files.keys():
+        for i, discharge in enumerate(reach.hydro_q):
+            qstr = str(discharge)
+            if qstr in reference_files.keys() or qstr in measure_files.keys():
                 cond = f"C{i+1}"
                 condition = ConditionConfig(
                     Discharge=discharge, Reference="", WithMeasure=""
                 )
-                if i < len(reference_files):
-                    condition.Reference = reference_files[discharge]
-                if i < len(measure_files):
-                    condition.WithMeasure = measure_files[discharge]
+                if qstr in reference_files.keys():
+                    condition.Reference = reference_files[qstr]
+                if qstr in measure_files.keys():
+                    condition.WithMeasure = measure_files[qstr]
 
                 config[cond] = condition.model_dump()
 
--- a/dfastmi/gui/dialog_view.py
+++ b/dfastmi/gui/dialog_view.py
@@ -192,16 +192,22 @@ class DialogView:
         self._update_condition_files()
 
     def _update_condition_files(self):
-        """Update the condition files."""
+        """
+        Update the condition files.
+        Use local copy of the dictionary because the callbacks from
+        the GUI updates will cause updates of the original dictionaries.
+        """
+        file_dictionary = self._view_model.reference_files.copy()
         for (
             condition_discharge,
             reference_file,
-        ) in self._view_model.reference_files.items():
+        ) in file_dictionary.items():
             self._update_condition_file_field(
                 reference_label, condition_discharge, reference_file
             )
 
-        for condition_discharge, measure_file in self._view_model.measure_files.items():
+        file_dictionary = self._view_model.measure_files.copy()
+        for ( condition_discharge, measure_file ) in file_dictionary.items():
             self._update_condition_file_field(
                 with_measure_label, condition_discharge, measure_file
             )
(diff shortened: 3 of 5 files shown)


dfastmi/__main__.py
```diff
@@ -94,30 +94,34 @@ def parse_arguments() -> Tuple[str, str, Optional[str], str, bool]:
...
+        choices=["UK", "NL"],
+        default="UK",
+        help=argparse.SUPPRESS,
     )
-    parser.set_defaults(language="UK")
 
     parser.add_argument(
-        "--mode", help="run mode 'BATCH', 'CLI' or 'GUI' (GUI is default)"
+        "--mode",
+        choices=["CLI", "BATCH", "GUI"],
```
MRVermeulenDeltares, line 105: Should CLI still be an option, since it is not noted in the help message?

dfastmi/gui/dialog_model.py
```diff
@@ -291,18 +291,17 @@ def _get_condition_configuration(
         measure_files: List,
     ) -> None:
         """Get condition configuration."""
-        num_files = min(len(reference_files), len(measure_files))
-
-        for i, discharge in enumerate(reach.hydro_q[:num_files]):
-            if discharge in reference_files.keys():
+        for i, discharge in enumerate(reach.hydro_q):
+            qstr = str(discharge)
+            if qstr in reference_files.keys() or qstr in measure_files.keys():
```
MRVermeulenDeltares, line 296: The reference_files and measure_files are handled like they are dictionaries, but the typehint suggest they are lists, which is correct?

dfastmi/gui/dialog_model.py
```diff
@@ -291,18 +291,17 @@ def _get_condition_configuration(
         measure_files: List,
     ) -> None:
         """Get condition configuration."""
-        num_files = min(len(reference_files), len(measure_files))
-
-        for i, discharge in enumerate(reach.hydro_q[:num_files]):
-            if discharge in reference_files.keys():
+        for i, discharge in enumerate(reach.hydro_q):
```
MRVermeulenDeltares, line 294: This will crash when the legacy mode is used.

dfastmi/gui/dialog_model.py
MRVermeulenDeltares: If it is a dictionary you can also make use of the .get method:
https://www.w3schools.com/python/ref_dictionary_get.asp
You can give it a default, it will return that value if it cannot find the key.
You can remove the two if statements and write it like this:

```
condition = ConditionConfig(
                    Discharge=discharge, 
                    Reference=reference_files.get(qstr, ""), 
                    WithMeasure=measure_files.get(qstr, "")
                )
```

If it cannot find the key it will return "" in this case.

dfastmi/__main__.py
hrajagers: As long as it's part of the code, I would keep it as option available. However, this option is no longer promoted. The description of the option is still available in the appendix of the technical reference manual, but no longer in the user manual.

By the way, it turned out that by listing it in the choices, the "CLI" option was still visible in the help message. Furthermore, it turned out that these choices are case sensitive by default. For these reasons, I've removed the choices option in a later commit and reverted to only a help text without mentioning the CLI mode.

dfastmi/gui/dialog_model.py
hrajagers: These variables may have been lists in the legacy code, but have switched to dictionaries a long time ago. I'm surprised that nothing complains about the mismatch with the type hints. I've used "mypy" in the past for checking type hints. I'm not sure whether that's the best option.

A quick check shows that mypy tool gives 356 warnings ... new issue created: [DFAST-214](https://issuetracker.deltares.nl/browse/DFAST-214) This list/dictionary issue doesn't appear in that list of warning, but it does include a warning concerning
`batch\FileNameRetrieverFactory.py:45: error: Need type annotation for "_creators" (hint: "_creators: dict[<type>, <type>] = ...")`
which may be related.

In dialog_view_model the type hints indicate that it is Dict[float, str] but the key could be an integer (0,1,2) for the legacy mode, a float representing the discharge in the standard mode, and a tuple of a float and a string (discharge and tidal boundary condition) when running in the research mode for the delta reaches of our rivers. Hence, this needs careful crafting.

hrajagers: Correction:

1. the mypy listing does include a warning on this topic
`dfastmi\gui\dialog_model.py:287: error: No overload variant of "__getitem__" of "list" matches argument type "str"  [call-overload]`
2. The discharge (float) is converted to str before using as key.

dfastmi/gui/dialog_model.py
hrajagers: Indeed. However, the GUI doesn't support legacy configurations (version 1 files). The batch mode and cli mode do support those, but we don't support them in the gui since the GUI logic would be completely different.

dfastmi/gui/dialog_model.py
hrajagers: Introduced a new type FilenameDict in dfastmi/kernel/typehints.py and used it in DialogModel and DialogViewModel. We'll probably need to use it at other locations as well.

dfastmi/gui/dialog_view.py
```diff
@@ -192,16 +192,22 @@ def _update_reach(self, data):
         self._update_condition_files()
 
     def _update_condition_files(self):
-        """Update the condition files."""
+        """
```
MRVermeulenDeltares, line 195: @ralph-peelen-deltares I would like to discuss these changes with you when you are available.

dfastmi/io/ConfigFileOperations.py
```diff
@@ -263,9 +263,12 @@ def _update_to_absolute_path(
...
         relative_path = config.get(section, key, fallback="")
-        relative_path_converted_to_absolute_path = str(
-            Path(rootdir).joinpath(Path(relative_path))
-        )
+        if len(relative_path) > 0:
+            relative_path_converted_to_absolute_path = str(
+                Path(rootdir).joinpath(Path(relative_path))
+            )
+        else:
+            relative_path_converted_to_absolute_path = ""
```
MRVermeulenDeltares, line 271: You can use `relative_path_converted_to_absolute_path = relative_path`
Since relative_path is already set to its fallback value of "" when the else is walked through.

dfastmi/gui/dialog_view.py
```diff
@@ -192,16 +192,22 @@ def _update_reach(self, data):
...
+        Update the condition files.
+        Use local copy of the dictionary because the callbacks from
+        the GUI updates will cause updates of the original dictionaries.
+        """
+        file_dictionary = self._view_model.reference_files.copy()
```
ralph-peelen-deltares, line 200: Please do not please intelligence in the View. We have setup a MVVM principle for the GUI. Which means the data is stored somehow in the model and the view displays calculated data in the GUI from the ViewModel. Please move this code to the ViewModel. Display the results in the View. If you need help or an example I can assist.

dfastmi/gui/dialog_model.py
ralph-peelen-deltares: Please do not use dictionaries. This breaks the principle of objected programming.

dfastmi/__main__.py
MRVermeulenDeltares: Discussed with Ralph resolving.

dfastmi/gui/dialog_model.py
MRVermeulenDeltares: Reviewed with Ralph, see #98

dfastmi/gui/dialog_model.py
MRVermeulenDeltares: Reviewed with Ralph, see #98.

dfastmi/gui/dialog_view.py
MRVermeulenDeltares: Discussed with Ralph, see PR #99

dfastmi/gui/dialog_view.py
MRVermeulenDeltares: See PR #100 and #101, no need to have a call to load configuration, this is part of viewmodel responsibility.
When instantiating the view, the data should be retrieved from the already instantiated viewmodel. With corrected eventing, blocking of emits is not needed.

tests/gui/test_dialog_model.py
```diff
@@ -56,13 +56,13 @@ def mock_rivers_object() -> MagicMock:
 @pytest.fixture
 def mock_reference_files() -> Dict[float, str]:
     """Fixture for creating mock reference files."""
-    return {80.1: "reference1.txt", 80.2: "reference2.txt"}
+    return {"80.1": "reference1.txt", "80.2": "reference2.txt"}
```
MRVermeulenDeltares, line 59: Returning prefered double because of new FilenameDict implementation.
see #98

dfastmi/io/ConfigFileOperations.py
MRVermeulenDeltares: see PR #106

dfastmi/gui/dialog_model.py
MRVermeulenDeltares: Discussed with Bert.

dfastmi/gui/dialog_model.py
MRVermeulenDeltares: Discussed with Bert.

dfastmi/gui/dialog_view.py
MRVermeulenDeltares: Discussed with Bert.

dfastmi/gui/dialog_view.py
MRVermeulenDeltares: Discussed with Bert.

dfastmi/io/ConfigFileOperations.py
MRVermeulenDeltares: Discussed with Bert.

tests/gui/test_dialog_model.py
MRVermeulenDeltares: Discussed with Bert.

PR #90, opened 2024-06-04
DFAST-206 Updated reporting of conditions

hrajagers:
* adjust reporting order of configuration settings
* make sure that conditions are always reported ... also for version 1 files.
* rename "Discharge [m3/s]" to "Condition" to prepare for tidal conditions


BuildScripts/BuildDfastmi_no_command_window.bat
```diff
@@ -1,43 +1,3 @@
...
- --include-data-files=dfastmi/open.png=dfastmi/open.png ^
- --include-data-files=docs/dfastmi_usermanual.pdf=dfastmi/dfastmi_usermanual.pdf ^
- --windows-force-stderr-spec=%PROGRAM%logs.txt ^
- --windows-force-stdout-spec=%PROGRAM%output.txt ^
- --windows-disable-console ^
- dfastmi
-
-rem end of build
\ No newline at end of file
+"%~dp0\BuildDfastmi.bat" --no-console
```
MRVermeulenDeltares, line 3: We should update this in teamcity buildstep for the [Build without command window](https://dpcbuild.deltares.nl/admin/editBuild.html?id=buildType:DFast_DFastMorphologicalImpact_BuildWithoutCommandWindow),  then this file can be removed.

dfastmi/Dutch_rivers_v3.ini
```diff
@@ -1,8 +1,8 @@
 [General]
-    Version    = 2.0
+    Version    = 3.0
```
MRVermeulenDeltares, line 2: What is the difference in version that it needs to be updated to v3?

dfastmi/batch/ConfigurationCheckerFactory.py
```diff
@@ -74,8 +74,7 @@ def generate(configuration_version: Version) -> AConfigurationCheckerBase:
         )
 
 
-legacy_version = Version("1.0")
-ConfigurationCheckerFactory.register_creator(legacy_version, ConfigurationCheckerLegacy)
+ConfigurationCheckerFactory.register_creator(Version("1.0"), ConfigurationCheckerLegacy)
```
MRVermeulenDeltares, line 77: do we want to remove the explicit mention of a legacy version?

dfastmi/batch/core.py
```diff
@@ -131,7 +131,7 @@ def batch_mode_core(
     with report_path.open(mode="w", encoding="utf-8") as report:
         _log_header(report)
 
-        cfg_version = _get_verion(rivers, config)
+        cfg_version = _get_version(rivers, config)
```
MRVermeulenDeltares, line 134: nice catch!

dfastmi/batch/core.py
```diff
@@ -216,46 +215,52 @@ def _report_analysis_settings_header(report: TextIO):
...
-        if section[0].lower() == "c":
-            discharge_file_name = config.get(section, "Discharge", fallback="")
-            reference_file_name = _get_file_name(config, section, "Reference")
-            measure_file_name = _get_file_name(config, section, "WithMeasure")
-            _report_analysis_conditions_values(
-                discharge_file_name, reference_file_name, measure_file_name, report
-            )
+        if section[0].lower() == "c" or section[0].lower() == "q":
+            discharge_str = config.get(section, "Discharge", fallback="")
+            discharge = float(discharge_str)
```
MRVermeulenDeltares, line 252: this will raise a valueerror if Discharge cannot be found in config

dfastmi/batch/core.py
```diff
@@ -216,46 +215,52 @@ def _report_analysis_settings_header(report: TextIO):
...
+            condition = "{:7.1f} m3/s".format(discharge)
+            if discharge <= q_threshold:
```
MRVermeulenDeltares, line 254: You can update this if else to make it more readable, e.g.: (I took the main version for the example, it has some more changes than suggested here)

```
        condition = "{:7.1f} m3/s".format(q)
        if q <= initialized_config.q_threshold:
            reference_file_name = _get_file_name(files[0])
            measure_file_name = _get_file_name(files[1])
            comment = ""
        elif key in filenames:
            reference_file_name = "---"
            measure_file_name = "---"
            comment = (measure not active)
        else:
            reference_file_name = "xxx"
            measure_file_name = "xxx"
            comment = (not specified)

        _report_analysis_conditions_values(
            condition, reference_file_name, measure_file_name, comment, report
        )
```

In this example I assumed the _report_analysis_conditions_values is also updated to include a comment, instead of sneaking in a comment in the measure_file_name.

In the main the _report_used_file_names method contains 2 big if elif else conditions, this could be split up in two seperate methods.

dfastmi/io/RiversObject.py
```diff
@@ -94,24 +94,24 @@ def _read_rivers_file(self, filename):
...
-        if iversion == 1:
+        if self.version == Version("1"):
             self._read_rivers_legacy(river_data)
 
-        else:  # iversion == 2
+        else:
             self._read_rivers(river_data)
 
-    def _parse_reaches(self, config: configparser.ConfigParser, iversion):
+    def _parse_reaches(self, config: configparser.ConfigParser, version: Version):
```
MRVermeulenDeltares, line 106: You can leave passing in the version argument out of this method call, since you pass in a class property you can just use it in that method. e.g. self.version.

dfastmi/batch/core.py
```diff
@@ -216,46 +215,52 @@ def _report_analysis_settings_header(report: TextIO):
...
     for section in config:
-        if section[0].lower() == "c":
-            discharge_file_name = config.get(section, "Discharge", fallback="")
-            reference_file_name = _get_file_name(config, section, "Reference")
-            measure_file_name = _get_file_name(config, section, "WithMeasure")
-            _report_analysis_conditions_values(
-                discharge_file_name, reference_file_name, measure_file_name, report
-            )
+        if section[0].lower() == "c" or section[0].lower() == "q":
+            discharge_str = config.get(section, "Discharge", fallback="")
```
MRVermeulenDeltares, line 251: Change into:

get --> getfloat

BuildScripts/BuildDfastmi_no_command_window.bat
MRVermeulenDeltares: Discussed with Ralph, keep as is.

dfastmi/batch/ConfigurationCheckerFactory.py
MRVermeulenDeltares: Discussed with Ralph, leave as is.

dfastmi/batch/core.py
MRVermeulenDeltares: should be resolved by suggestion mentioned above.

dfastmi/batch/core.py
MRVermeulenDeltares: Resolved in PR #102, comment is now part of the message, instead of added onto the measure files.

dfastmi/io/RiversObject.py
MRVermeulenDeltares: @ralph-peelen-deltares  will create PR.

dfastmi/batch/core.py
MRVermeulenDeltares: seems to have already been resolved in #94 with the use of initialized_config.discharges which is a vector with floats, so no need for casting to a float or retrieving from the config.

MRVermeulenDeltares: Discussed with Ralph, resolving conversation.

dfastmi/Dutch_rivers_v3.ini
MRVermeulenDeltares: @ralph-peelen-deltares 
Version 3 is identical to version 2, except for the version number.
It was decided to change this to 3 to keep the file version in line with the application version.

dfastmi/batch/core.py
MRVermeulenDeltares: Discussed with Bert.

dfastmi/io/RiversObject.py
MRVermeulenDeltares: PR:#108 Discussed with Bert.

dfastmi/Dutch_rivers_v3.ini
MRVermeulenDeltares: Discussed with Ralph.

PR #91, opened 2024-06-05
DFAST-209: Appropriate Capitalization of config files, solve small GUI flashing, load config upon start

hrajagers:
* Specifying a configuration file on the command line didn't actually load the file. Loading the config_file moved from DialogModel to DialogView. The file is loaded only at the time of the activation of the user interface since only then are all signals active that guarantee that all fields of the file are actually loaded.
* The ActivationChange event is triggered whenever the dialog gains or loses focus, e.g. when the dialog is created (that's the one that we're interested in), deleted (at the end) and whenever a file/folder open dialog appears and disappears.
* Don't trigger callback events when setting the branch and reach texts from the _update_branch and _update_reach calls since those will trigger these routines again.
* Associate all newly created widgets with the main window. Don't pass self._win if self is also passed. Increases interface sizes without reason.
* Set visibility of widgets to false before deleting them ... if you don't then a small dummy dialog will flash on the screen for each deleted widget.
* Check whether all sections and options(i.e. keywords) have appropriate capitalization in the loaded configuration file (not yet checking the Q/C* sections/keywords).
* Update the associated tests.


dfastmi/gui/dialog_view.py
```diff
@@ -163,7 +184,9 @@ def _update_branch(self, data):
         self._case_description.setText(self._view_model.model.case_description)
 
         # Update branch and reach selection
+        self._branch.blockSignals(True)
```
MRVermeulenDeltares, line 187: Maybe move this to a method, since it used multiple times in a similar way, also for reach.
e.g.:
```
self.set_combobox_label_without_signals(self._branch, data)
self.set_combobox_label_without_signals(self._reach, data)

def set_combobox_label_without_signals(self, combobox : QComboBox, text : str):
        combobox.blockSignals(True)
        combobox.setCurrentText(text)
        combobox.blockSignals(False)
```

dfastmi/io/ConfigFileOperations.py
```diff
@@ -205,6 +206,68 @@ def load_configuration_file(filename: str) -> configparser.ConfigParser:
         rootdir = os.path.dirname(filename)
         return ConfigFileOperations._config_to_absolute_paths(rootdir, config)
 
+    @staticmethod
+    def _config_case_check(
+        config: configparser.ConfigParser,
+    ) -> configparser.ConfigParser:
+        config = ConfigFileOperations._config_case_check_section(config, "General")
+        for key in [
```
MRVermeulenDeltares, line 214: Make use of the generalconfig class which uses pydantic to do this check automatically.

dfastmi/gui/dialog_model.py
```diff
@@ -71,38 +72,17 @@ class DialogModel:
...
         Initialize the DialogModel.
 
         Args:
             rivers_configuration (RiversObject): Configuration of rivers.
-            config_file (Optional[str], optional): Configuration file. Defaults to None.
         """
         self.rivers = rivers_configuration
 
-        if config_file:
-            self.load_configuration(config_file)
```
MRVermeulenDeltares, line 87: Is the load_configuration still used or can it be fully removed now?

tests/gui/test_dialog_model.py
```diff
@@ -224,17 +222,13 @@ def test_load_configuration_init(mock_rivers_object: MagicMock, mocker) -> None:
...
-    filename = "test.cfg"
     config = ConfigParser()
     config["General"] = {}
-    mock_load_config = mocker.patch.object(
-        ConfigFileOperations, "load_configuration_file", return_value=config
-    )
-    model = DialogModel(mock_rivers_object, filename)
-    mock_load_config.assert_called_once_with(filename)
+    model = DialogModel(mock_rivers_object)
+
```
MRVermeulenDeltares, line 231: The name of this tests should be updated, since you are not loading the config anymore.

dfastmi/gui/dialog_view.py
```diff
@@ -933,31 +955,29 @@ def _menu_open_manual(self):
         ---------
         None
         """
-        subprocess.Popen(self._view_model.manual_filename, shell=True)
+        os.startfile(self._view_model.manual_filename)
```
ralph-peelen-deltares, line 958: good improvement.
**Windows Only**: Use os.startfile(self._view_model.manual_filename) for simplicity and safety when you just need to open a file with its default application.

dfastmi/gui/dialog_model.py
MRVermeulenDeltares: Discussed with Ralph, can be closed.

dfastmi/gui/dialog_view.py
MRVermeulenDeltares: Corrected eventing in #99 and #101

dfastmi/gui/dialog_view.py
MRVermeulenDeltares: Discussed with Marlon, can be closed.

dfastmi/io/ConfigFileOperations.py
MRVermeulenDeltares: New issue to be created.

tests/gui/test_dialog_model.py
MRVermeulenDeltares: Test can be removed. @MRVermeulenDeltares create PR.

dfastmi/io/ConfigFileOperations.py
MRVermeulenDeltares: https://issuetracker.deltares.nl/browse/DFAST-216

tests/gui/test_dialog_model.py
MRVermeulenDeltares: See PR #107

dfastmi/gui/dialog_view.py
MRVermeulenDeltares: Discussed with Bert.

dfastmi/io/ConfigFileOperations.py
MRVermeulenDeltares: Discussed with Bert.

tests/gui/test_dialog_model.py
MRVermeulenDeltares: Discussed with Bert.

PR #93, opened 2024-06-05
D-FAST MI failed to load gui without proper config file on command line

hrajagers:
As indicated by the title ... this wasn't caught by the automated tests ... only by the manual testing as described in the tech ref

### DIFF
--- a/dfastmi/gui/dialog_view_model.py
+++ b/dfastmi/gui/dialog_view_model.py
@@ -359,8 +359,15 @@ class DialogViewModel(QObject):
                 self._reference_files[cond_discharge] = section.get("Reference", "")
                 self._measure_files[cond_discharge] = section.get("WithMeasure", "")
 
-        self.current_branch = self.model.rivers.get_branch(self.model.branch_name)
-        self.current_reach = self.current_branch.get_reach(self.model.reach_name)
+        branch = self.model.rivers.get_branch(self.model.branch_name)
+        if not branch: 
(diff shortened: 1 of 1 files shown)


MRVermeulenDeltares (commented):
The GUI not starting was not caught by test test_basic_gui in test_basic.py?

dfastmi/gui/dialog_view_model.py
```diff
@@ -359,8 +359,15 @@ def load_configuration(self, filename: str) -> bool:
                 self._reference_files[cond_discharge] = section.get("Reference", "")
                 self._measure_files[cond_discharge] = section.get("WithMeasure", "")
 
-        self.current_branch = self.model.rivers.get_branch(self.model.branch_name)
-        self.current_reach = self.current_branch.get_reach(self.model.reach_name)
+        branch = self.model.rivers.get_branch(self.model.branch_name)
```
ralph-peelen-deltares, line 362: why do you want to set a river branch and reach which is not in the analysis configuration file, would None be sufficient and show nothing in the GUI?

hrajagers: The default (internal) config definition doesn't have a branch and reach set. The branch and reach are dropdown fields, so they can't be None/empty. I'm perfectly fine with defaulting to the first entry in the respective lists and that's the purpose of this code.

MRVermeulenDeltares: Discussed with Ralph.

PR #94, opened 2024-06-06
Various bug fixes mostly related to interactive editing and running of cases via GUI

hrajagers:


dfastmi/batch/core.py
```diff
@@ -245,26 +245,41 @@ def _report_analysis_conditions_header(report: TextIO):
...
+    for i, q in enumerate(initialized_config.discharges):
+        if not q:  # should only happen for version 1 files
+            continue
+        if 0 in filenames:
+            key = i
+        elif initialized_config.needs_tide:
+            key = (q, initialized_config.tide_bc[i])
+        else:
+            key = q
+        condition = "{:7.1f} m3/s".format(q)
```
MRVermeulenDeltares, line 264: Suggestion to move this to a descriptable method and refactor a litte bit, see example:
```
        condition, reference_file_name, measure_file_name, comment = get_analysis_condition_values_for_logging(initialized_config, filenames, q, key)

        _report_analysis_conditions_values(
            condition, reference_file_name, measure_file_name, comment, report
        )


def get_analysis_condition_values_for_logging(initialized_config, filenames, q, key):
    condition = "{:7.1f} m3/s".format(q)
    
    if q <= initialized_config.q_threshold:
        reference_file_name = "---"
        measure_file_name = "---"
        comment = "(measure not active)"
    elif key in filenames:
        files = filenames[key]
        reference_file_name = _get_file_name(files[0])
        measure_file_name = _get_file_name(files[1])
        comment = ""
    else:
        reference_file_name = "xxx"
        measure_file_name = "xxx"
        comment = "(not specified)"
        
    return condition, reference_file_name, measure_file_name, comment
```

This example moves the behviour that decides the data for logging to a method.
The actual message in the UK and NL messages files should also be updated to include comment, as the _report_analysis_conditions_values can have the comment inserted seperately and wont have it merged into the measure message.

dfastmi/gui/dialog_model.py
```diff
@@ -205,14 +206,15 @@ def check_configuration(
         config = self.get_configuration(branch, reach, reference_files, measure_files)
         return check_configuration(self.rivers, config)
 
-    def run_analysis(self) -> bool:
+    def run_analysis(self, gui: bool = False) -> bool:
```
ralph-peelen-deltares, line 209: is this really necessary? why? This is part of a MVVM principle,  why wouldn't you use dfastmi.batch.core.batch_mode_core directly

ralph-peelen-deltares: It is also not used in the code.. please remove

dfastmi/gui/dialog_model.py
```diff
@@ -221,12 +223,9 @@ def run_analysis(self) -> bool:
             We call batch_mode_core which can throw and log an exception.
             If thrown, analysis has failed.
         """
-        try:
-            success = dfastmi.batch.core.batch_mode_core(
-                self.rivers, False, self.config
-            )
-        except:
-            success = False
+        success = dfastmi.batch.core.batch_mode_core(
```
ralph-peelen-deltares, line 226: We always, provide default = false for GUI parameter, no need to pass through

dfastmi/gui/dialog_view_model.py
```diff
@@ -110,10 +112,32 @@ def current_reach(self, value):
...
+    @qthreshold.setter
+    def qthreshold(self, value: float):
+        """
+        Setter for the current threshold discharge.
+        After set notify the view of the change.
+
+        Arguments:
+            value (float): The threshold discharge to set.
+        """
+        self._qthreshold = value
```
ralph-peelen-deltares, line 134: Please check if new qthreshhold is already same as the one currently in view_model, no need to emit again.
```
if self_qthreshold == value:
    return
    
```

dfastmi/gui/dialog_model.py
ralph-peelen-deltares: or we need to provide always **True** here...

dfastmi/gui/dialog_view_model.py
```diff
@@ -286,8 +311,8 @@ def _update_qvalues(self) -> None:
             if self.current_reach.auto_time:
                 _, time_mi = ConfigurationInitializer.set_times(
                     self._current_reach.hydro_q,
```
ralph-peelen-deltares, line 313: Why using field instead of property?

tests/gui/test_dialog_model.py
```diff
@@ -128,30 +128,6 @@ def test_run_analysis_failure(mock_batch_mode_core: MagicMock) -> None:
     assert result is False
 
 
-def test_run_analysis_exception(mock_batch_mode_core: MagicMock) -> None:
```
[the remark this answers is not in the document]
ralph-peelen-deltares: maybe rewrite to:
```
def test_run_analysis_exception(mock_batch_mode_core: MagicMock) -> None:
    """
    Test case for exception during analysis run.
    given: A DialogModel instance.
    when: Calling the run_analysis method with batch_mode_core raising an exception.
    then: The batch_mode_core function is called and the return value of run_analysis is False (failure).
    """
    # Set the side effect of batch_mode_core to raise an exception
    mock_batch_mode_core.side_effect = Exception("mocked")

    # Create an instance of DialogModel
    model = DialogModel(rivers_configuration=None)

    # Call the run_analysis method
    try:
        model.run_analysis()
    except Exception as e:
        assert "mocked" in traceback.format_exc()
        assert str(e) == "mocked"               

    # Check that batch_mode_core was called
    mock_batch_mode_core.assert_called_once()    
```

dfastmi/batch/core.py
MRVermeulenDeltares: Resolved see PR #102

dfastmi/io/ConfigFileOperations.py
```diff
@@ -153,15 +153,16 @@ def _update_to_relative_path(
...
         absolute_path = config.get(section, key, fallback="")
-        if len(absolute_path) == 0:
-            absolute_path = rootdir
-
-        absolute_path_converted_to_relative_path = str(
-            os.path.relpath(absolute_path, rootdir)
-        )
 
-        if absolute_path_converted_to_relative_path == ".":
+        if len(absolute_path) == 0:
```
MRVermeulenDeltares, line 157: Why was this updated, was something not working?

dfastmi/gui/dialog_view.py
```diff
@@ -713,34 +722,29 @@ def _create_case_input(self, layout: QBoxLayout) -> None:
...
-    def _update_qthreshold(self) -> None:
+    def _updated_qthreshold(self) -> None:
         """
         Update discharge threshold.
 
         Returns:
             None
         """
         if self._qthr.hasAcceptableInput():
+            new_qthreshold = max(
```
ralph-peelen-deltares, line 742: please update _qthr.setText(new_value) with correct value then!

ralph-peelen-deltares: see pr #105

dfastmi/batch/core.py
MRVermeulenDeltares: Discussed with Bert.

dfastmi/gui/dialog_model.py
MRVermeulenDeltares: This is added in relation to the generated figures.
When all figure are generated, this parameter is used to determine if the control of closing the figures should be returned to the user.

dfastmi/gui/dialog_model.py
MRVermeulenDeltares: See comment above.

dfastmi/gui/dialog_view_model.py
MRVermeulenDeltares: Discussed with Bert.

dfastmi/gui/dialog_view_model.py
MRVermeulenDeltares: @ralph-peelen-deltares, Would you be able to discuss this shortly with Bert and explain it further.

dfastmi/io/ConfigFileOperations.py
MRVermeulenDeltares: Behaviour with empty paths is currently unsure.
@hrajagers will have a look at this.

tests/gui/test_dialog_model.py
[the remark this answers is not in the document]
MRVermeulenDeltares: Discussed with Bert.

dfastmi/gui/dialog_model.py
MRVermeulenDeltares: Discussed with Ralph.

MRVermeulenDeltares: See #103

dfastmi/gui/dialog_model.py
MRVermeulenDeltares: See PR #103, Discussed with Ralph.

dfastmi/gui/dialog_view_model.py
hrajagers: I noticed an inconsistency between using the "_current_reach" field and in the line above and the "current_reach" property in the lines below. I decided to make these consistent. Since it's inside the objects own code, I thought that using the fields would be fine. If the property would be doing something special the we would be doing it three times ... or we should first get the current_reach and then the hydro_q, qfit and qstagnant properties of that item.

dfastmi/io/ConfigFileOperations.py
hrajagers: The main issue that I encountered was that the absolute_path to relative_path worked only when the path and the reference are on the same drive. It would crash the code when trying to come up with a relative path to a file on a different drive. The new code falls back to using the absolute path when that happens.

Furthermore, if the absolute_path (this is the path shown in the GUI) is empty, the relative path should also be empty. Otherwise, fields that haven't been entered will suddenly point to the rootdir. This was fixed by the line that converts relative paths "." to empty string again at the end, but this would not allow you to use the rootdir/current folder as the folder for output.

MRVermeulenDeltares: We will leave it as is.

PR #98, opened 2024-06-12
Code changes associated with reviews of pull requests 89-94

hrajagers:


tests/gui/test_dialog_model.py
```diff
@@ -271,7 +281,7 @@ def test_load_configuration_except_true(dialog_model: DialogModel, mocker) -> No
 def test_check_configuration(
     dialog_model: DialogModel,
     mock_branch: MagicMock,
-    mock_reach: MagicMock,
+    mock_areach: MagicMock,
     mock_reference_files: Dict[float, str],
     mock_measure_files: Dict[float, str],
```
hrajagers, line 286: I expect that these should be filenameDict as well.

dfastmi/kernel/typehints.py
```diff
@@ -14,3 +14,6 @@
 
 # Represents a list of booleans of variable lengths.
 BoolVector = List[bool]
+
+# Filename dictionary
+FilenameDict = Dict[Union[int, str, Tuple[float, str]], str]
```
hrajagers, line 19: Note that the FilenameDict uses a Union for the dictionary key:
* int: for backward compatibility with 0,1,2 used for Version 1 configurations
* str: for discharges ... I see that you've changed those to floats which shouldn't be allowed according to this definiiton
* Tuple[float, str]: for pairs of discharge, tidal_boundary conditions.
I was in doubt whether to represent the discharges as float, or str. Since we treat tidal boundary conditions as str, I thought it might be more consistent to store discharges also as str. However, the current type definition is inconsistent for the discharge only key and the discharge-tide Tuple. Given your changes, it's probably better to change it to Union[int, float, Tuple[float, str]]

hrajagers: Furthermore, #99 introduces Path as value instead of str.

MRVermeulenDeltares: @ralph-peelen-deltares, we need to check this together with #89

tests/gui/test_dialog_model.py
MRVermeulenDeltares: @ralph-peelen-deltares we need to check this.

MRVermeulenDeltares: Will update

MRVermeulenDeltares: done

dfastmi/kernel/typehints.py
MRVermeulenDeltares: To be resolved in #99

PR #99, opened 2024-06-17
DFAST-212 resolve eventing

MRVermeulenDeltares:


dfastmi/gui/dialog_view_model.py
```diff
@@ -53,8 +54,10 @@ class DialogViewModel(QObject):
     save_plot_changed = pyqtSignal(bool)
     figure_dir_changed = pyqtSignal(str)
     output_dir_changed = pyqtSignal(str)
-    _reference_files: Dict[float, str] = {}
-    _measure_files: Dict[float, str] = {}
+    reference_files_changed = pyqtSignal(str, float, Path)
+    measure_files_changed = pyqtSignal(str, float, Path)
+    _reference_files: Dict[float, Path] = {}
+    _measure_files: Dict[float, Path] = {}
```
hrajagers, line 60: Link to #98 ... This should be FilenameDict, correct?

hrajagers: Ah, we should then change the definition of FilenameDict since that refers to str instead of Path.

dfastmi/gui/dialog_view_model.py
```diff
@@ -145,16 +148,16 @@ def qthreshold(self, value: float):
...
+        Dict[float, Path]: The reference files.
         """
         return self._reference_files
 
     @property
-    def measure_files(self) -> Dict[float, str]:
+    def measure_files(self) -> Dict[float, Path]:
         """
-        Dict[float, str]: The measurement files.
+        Dict[float, Path]: The measurement files.
```
hrajagers, line 160: same comment as above.

hrajagers (commented):
Need to check consistency of pull requests #98 and #99.

dfastmi/gui/dialog_view_model.py
MRVermeulenDeltares: Discussed with Ralph, we will set it back to str.

PR #100, opened 2024-06-17
DFAST-212 load config file data correctly when creating dialog view

MRVermeulenDeltares:


hrajagers:
Changed code looks simpler. I haven't tested it yet. I don't think that we have an automated or manual test defined to verify whether a config specified on the command line is loaded properly in the GUI.

PR #101, opened 2024-06-17
DFAST-212 remove signal blocking only emit to view when reach or branch is really changed

MRVermeulenDeltares:


hrajagers:
OK, instead of suppressing signals, you include a check to see whether value has changed and hence if further action is needed.

PR #102, opened 2024-06-17
DFAST-212: Review updated reporting of conditions

MRVermeulenDeltares:


dfastmi/batch/core.py
```diff
@@ -260,22 +260,38 @@ def _report_used_file_names(
...
             )
+        )
+
+        _report_analysis_conditions_values(
+            condition, reference_file_name, measure_file_name, comment, report
+        )
+
+
+def get_analysis_condition_values_for_logging(
+    initialized_config: AConfigurationInitializerBase, filenames, q, key
```
hrajagers, line 276: Add typing

hrajagers:
Agree with this change. It's cleaner. We may have to change this code again once the tool gets used because I'm not sure that reporting just the file names is enough. It's likely that all filenames will be equal in daily practice and only the full paths are different. However, that's for later concern.

dfastmi/batch/core.py
MRVermeulenDeltares: Will do!

MRVermeulenDeltares: Done

PR #106, opened 2024-06-19
DFAST-212 update _update_to_absolute_path with review suggestion.

MRVermeulenDeltares:


hrajagers:
OK, to split _update_to_absolute_path.
The new function _get_absolute_path_from_relative_path could be useful beyond config, but moving it to a separate package will add to package interdependencies.
I don't see the benefit in returning relative_path when len(relative_path) == 0. I agree that in that case relative_path equals "", so the result is largely the same.

PR #107, opened 2024-06-19
DFAST-212 remove unneeded test.

MRVermeulenDeltares:


hrajagers:
Agree that this test here isn't appropriate anymore. Do we need an extra test somewhere else?

PR #109, opened 2024-06-21
DFAST-212 : When selecting a reach calulate sedimentation length from the reach discharge threshold

ralph-peelen-deltares:
- Do this in the view_model
- Emit on change to view

### DIFF
--- a/dfastmi/gui/dialog_view_model.py
+++ b/dfastmi/gui/dialog_view_model.py
@@ -51,6 +51,7 @@ class DialogViewModel(QObject):
     branch_changed = pyqtSignal(str)
     reach_changed = pyqtSignal(str)
     qthreshold_changed = pyqtSignal(float)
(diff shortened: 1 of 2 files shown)


dfastmi/gui/dialog_view_model.py
```diff
@@ -367,6 +369,8 @@ def _update_slength(self) -> None:
         except:
             self.slength = "---"
 
+        self.slength_changed.emit(self.slength)
```
MRVermeulenDeltares, line 372: ain't it better to make slength a property and put the emit in the setter of the property?

Similar as with branch and reach.

MRVermeulenDeltares (approved):

dfastmi/gui/dialog_view_model.py
ralph-peelen-deltares: It cannot be edited in the GUI because it is a calculated value.

PR #112, opened 2024-06-26
DFAST-220: QThreshold is not correct in GUI after new selection of reach

ralph-peelen-deltares:
- set min expected qthreshold (reach stagnant discharge) in qthreshold label
- make caching possible for qthreshold during application runtime usage
- take in account qthreshold value loaded from app configuration, cache this too

### DIFF
--- a/dfastmi/gui/dialog_model.py
+++ b/dfastmi/gui/dialog_model.py
@@ -116,20 +116,10 @@ class DialogModel:
         """Get Qthreshold."""
         return self.section.getfloat("Qthreshold", 0.0)
 
-    @qthreshold.setter
-    def qthreshold(self, value: float):
-        """Set Qthreshold."""
-        self.section["Qthreshold"] = str(value)
-
     @property
     def ucritical(self) -> float:
         """Get Ucritical."""
         return self.section.getfloat("Ucrit", 0.3)
-
-    @ucritical.setter
-    def ucritical(self, value: float):
-        """Set Ucritical."""
-        self.section["Ucrit"] = str(value)
 
     @property
     def output_dir(self) -> str:
@@ -210,9 +200,32 @@ class DialogModel:
         reach: AReach,
         reference_files: FilenameDict,
         measure_files: FilenameDict,
+        ucritical : float,
+        qthreshold : float
     ) -> bool:
-        """Check configuration."""
-        config = self.get_configuration(branch, reach, reference_files, measure_files)
+        """
+        Check if configuration can be created.
+
+        Arguments
+        ---------
+        branch: Branch
+            Selected branch which is used and should be in this config.
+        reach: AReach
+            Selected reach which is used and should be in this config.
+        reference_files: FilenameDict
+            Selected reference files which is used and should be in this config.
+        measure_files: FilenameDict
+            Selected measure files which is used and should be in this config.
+        ucritical : float
+            Selected minimal critical flow value which is used and should be in this config.
+        qthreshold : float
+            Selected discharge threshold value which is used and should be in this config.
+
+        Returns
+        -------
+            Boolean indicating whether the (D-FAST MI analysis) configuration can be created.
+        """
+        config = self.get_configuration(branch, reach, reference_files, measure_files, ucritical, qthreshold)
         return check_configuration(self.rivers, config)
 
     def get_configuration(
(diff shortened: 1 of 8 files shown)


dfastmi/gui/dialog_view.py
```diff
@@ -651,9 +657,14 @@ def _create_qthreshhold_input(
         self._qthr.setValidator(double_validator)
         self._qthr.editingFinished.connect(self._updated_qthreshold)
         self._qthr.setToolTip(gui_text("qthr_tooltip"))
-        qthrtxt = QLabel(gui_text("qthr"), self._win)
+
+        qthreshold_label_text = gui_text(
+            "qthr",
+            placeholder_dictionary={"stagnant": str(self._view_model.model.qthreshold)},
```
MRVermeulenDeltares, line 663: you want to retrieve it from your viewmodel correct?
and your viewmodel will retrieve it from your model right?

ralph-peelen-deltares: ai good catch, euh... need to think why i didn't get the view_model one

dfastmi/gui/dialog_view.py
```diff
@@ -224,17 +226,21 @@ def _update_ucritical(self, ucrit: float, default: float):
...
-            data: The dischage threshold.
+            qthreshold: The dischage threshold.
+            qstagnant: The dischage stagnant threshold.
         """
         # Update the threshold discharge in the GUI
-        self._qthr.setText(str(data))
+        self._qthr.setText(str(qthreshold))
 
         # Update labels and text fields
+        self._qthrtxt.setText(
```
MRVermeulenDeltares, line 241: Maybe give gthrtxt a different name, now its a little bit confusing with the _qthr.setText and the _qthrtxt.setText

The _qthr is the input field right?

and the _qthrtxt is the label infront of the input field which is updated with the stagnant value right?

dfastmi/gui/dialog_view.py
ralph-peelen-deltares: I look into all .model properties if they are needed. plotting flags are also like this...

dfastmi/gui/dialog_view.py
ralph-peelen-deltares: yes. but was already name locally like this. To avoid confision will rename to _qtrlabel

dfastmi/gui/dialog_view_model.py
```diff
@@ -350,9 +352,14 @@ def _initialize_qthreshold(self):
...
-            self.model.qthreshold = self.current_reach.qstagnant
-        self.qthreshold = self.model.qthreshold
+        self._qthreshold = 0.0
+        if (self.current_branch, self.current_reach) in self._qthreshold_cache:
+            self.qthreshold = max(
+                self._qthreshold_cache[(self.current_branch, self.current_reach)],
+                self.current_reach.qstagnant,
+            )
+        else:
+            self.qthreshold = self.current_reach.qstagnant
```
MRVermeulenDeltares, line 362: so the model shouldn't be updated anymore?

ralph-peelen-deltares: It does that in the property

dfastmi/gui/dialog_view_model.py
```diff
@@ -179,10 +180,11 @@ def qthreshold(self, value: float):
         value = max(value, self.current_reach.qstagnant)
         self._qthreshold = value
         self.model.qthreshold = value
+        self._qthreshold_cache[(self.current_branch, self.current_reach)] = value
 
         self._update_slength()
         # Notify the view of the change
-        self.qthreshold_changed.emit(self._qthreshold)
+        self.qthreshold_changed.emit(self.qthreshold, self.current_reach.qstagnant)
```
MRVermeulenDeltares, line 187: shouldn't qstagnant have its own property and emit if it is needed?

ralph-peelen-deltares: no

MRVermeulenDeltares: isn't it now just piggybacking on the qthreshold emit?

MRVermeulenDeltares (commented):
See review comments.

dfastmi/gui/dialog_view_model.py
ralph-peelen-deltares: qstagnant is the discharge stagnant of a reach (minimum/'drempel'), threshold can be higher.

dfastmi/gui/dialog_view_model.py
MRVermeulenDeltares: Understood resolving converstation.

dfastmi/gui/dialog_model.py
```diff
@@ -116,21 +116,11 @@ def qthreshold(self) -> float:
         """Get Qthreshold."""
         return self.section.getfloat("Qthreshold", 0.0)
 
-    @qthreshold.setter
-    def qthreshold(self, value: float):
-        """Set Qthreshold."""
-        self.section["Qthreshold"] = str(value)
-
     @property
     def ucritical(self) -> float:
```
MRVermeulenDeltares, line 120: The viewmodel now only makes use of the readonly qthreshold and ucritical when loading a config, in all other cases it will use the view_model version of these properties.

And these propeties will only be update/set in the model when you get a configuration.

ralph-peelen-deltares: yes this is expected.

MRVermeulenDeltares: Discussed, this is correct.

MRVermeulenDeltares (approved):

PR #113, opened 2024-06-30
Use an integrated PDF viewer to avoid the security risk of a system call

hrajagers:
Sigrid associates a security risk with the `os.startfile` call.
Using an integrated PDF viewer using Qt6 might resolve this issue.


hrajagers:
We may look at this branch when upgrading from PyQt5 to PySide6, but it seems that we don't need the internal PDF viewer after the merge of #116 which resolves the security issue identified by Sigrid.

PR #120, opened 2024-07-09
DFAST-230: Update the header/introduction text

hrajagers:
Update the header/introduction text when D-FAST MI runs using version 2.0/3.0 configuration files (always when using the GUI, and in most cases when running in batch mode).

When using legacy version 1.0 files, the algorithm is equivalent to that of the original WAQMORF code (just with an option to use D-Flow FM results).

Also update the corresponding introduction paragraph in the manuals.

### DIFF
--- a/dfastmi/batch/core.py
+++ b/dfastmi/batch/core.py
@@ -130,9 +130,8 @@ def batch_mode_core(
     )
 
     with report_path.open(mode="w", encoding="utf-8") as report:
-        _log_header(report)
-
         cfg_version = _get_version(rivers, config)
+        _log_header(report, cfg_version)
 
         branch_name = config.get("General", "Branch", fallback="")
         branch = rivers.get_branch(branch_name)
@@ -406,7 +405,7 @@ def _get_version(rivers: RiversObject, config: ConfigParser) -> Version:
     return cfg_version
 
 
-def _log_header(report: TextIO) -> None:
+def _log_header(report: TextIO, cfg_version: Version) -> None:
     """
     Will log into the report default static header information
 
@@ -420,8 +419,12 @@ def _log_header(report: TextIO) -> None:
     None
     """
     prog_version = dfastmi.__version__
+    if cfg_version == Version("1.0"):
+        header = "header_legacy"
+    else:
+        header = "header"
     ApplicationSettingsHelper.log_text(
-        "header", dict={"version": prog_version}, file=report
+        header, dict={"version": prog_version}, file=report
     )
     ApplicationSettingsHelper.log_text("limits", file=report)
     _report_section_break(report)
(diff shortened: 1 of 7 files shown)


hrajagers:
@ralph-peelen-deltares Would you be able to look at this change? This is more than a textual change since the header text should be different for the legacy files (version 1.0) and the new files (version 3.0, and 2.0). The cli code now refers to the "header_legacy". The batch code prints the "header" or the "header_legacy" depending on the version of the configuration file.

I've considered adding unit tests to verify that "_log_header" writes the correct text, but that would only add to the number of places where the same text  is committed to the repository, and time would be wasted if we're going to remove the legacy mode in an one of the next releases anyway. So, I wouldn't do that.

hrajagers:
Sorry ... didn't create an issue for this ...

ralph-peelen-deltares (approved):
Solution is sufficient and inline with other reporting.

Although I would prefer to see a factory to separate versioning concerns in a single responsibility function and use an abstraction to create reporting so we don't need to check with in the function which version we are dealing with and what should be in the files, that would be a big refactoring for the reporting feedback module.

Would imply an abstract report object within core. The reporting object would have concrete implementations for legacy  (AnalyserAndReporterWaqua.analyse_and_report_waqua) and current way of reporting (AnalyserAndReporterDflowfm.analyse_and_report_dflowfm). In this concrete implementation the header would be written.. currently it is not correct and less maintainable. This is a risk. I love to discuss/orate further but I feel this is not inscope for this issue.

Maybe also create legacy messages.nl/uk.ini to keep this more maintainable (smaller files are easier to understand).

In the current implementation we already read the configuration version. We also haven't split the responsibility in the core.py python class in an abstract object. In _analyse_and_report we do a similar switch on versioning.

For current scope, let's keep this implementation. Let's please consider a new improvement for maintainability of reporting. If possible create a follow-up issue.

ralph-peelen-deltares:
To validate this change no unit tests are needed because it needs an integration test to write a report. Two integration test validate the header_legacy usage (in English). This is sufficient, change of language reporting is not inscope. Issue can be merged

hrajagers:
Hi @ralph-peelen-deltares , Thank you for the review. I share your desire to refactor the code. Maybe I would propose a slightly different approach that you sketch, or maybe I phrase it differently. I would probably try to split the workflow immediately after reading the configuration files, and have separate analysis and reporting workflows for the three options (legacy WAQUA, legacy FM, modern FM) ... a factory may indeed be a good way to do that. Those separate workflow should call shared routines whenever possible in as big chunks as possible, but only for really shared code that doesn't need to distinguish between the options. I think the duplicated code can very limited. This is easier to understand than having one workflow focusing on the shared elements that splits whenever necessary. However, there is currently no broader priority to start on such major refactoring right now, and therefore I've created a new issue DFAST-231 for that.

PR #133, opened 2025-03-31
move read xyc to dfastio repo and import it.

jeroen-deltares:
### read_xyc method de-duplication
The read_xyc method is used in both morphological impact and bankerosion of D-FAST. Using a commons library we can de-duplicate this method and increase maintainability.

**What was done:**
- Move the read_xyc method from this repo to a shared repo.
- Move the pytests for read_xyc to the shared repo.
- Remove the read_xyc method and the pytests from morphological impact.
- Add the commons library as a dependency for morphological impact.


MAfarrag:
tests were failing because of a `floating point error`, so i changed the comparison of the netcdf variables to use numpy and the comparison function now takes a tolerance value.

tests/batch/test_batch_core.py
```diff
@@ -196,6 +205,7 @@ def test_batch_mode_05(self):
         """
         Same as test_batch_mode_04 but includes centreline snapping.
         """
+        print("ssss")
```
hrajagers, line 208: Remove debug statement

tests/batch/test_batch_core.py
```diff
@@ -35,13 +36,21 @@ def compare_text_files(dir1, dir2, file1, file2=None, prefixes=None):
...
+        refdat = ncRef.variables[f][...]
+
+        if not np.allclose(result, refdat, atol=tol, equal_nan=True):
+            diff = np.abs(result - refdat)
+            max_diff = np.nanmax(diff)
+            raise AssertionError(
+                f"Field '{f}' differs between files. Max difference: {max_diff} exceeds tolerance {tol}."
+            )
+    ncRes.close()
+    ncRef.close()
```
hrajagers, line 53: I know that testing for equality of floating point values is bad practice, but that doesn't always hold for regression testing. It depends on how much numerical fluctuations you expect. In a complex simulation like D-Flow FM runs you may expect significant variability, but here I expect none or barely anything. Therefore I would have expected the use of np.array_equal instead of np.allclose. If we're going to use np.allclose, then I don't understand why we would lower the absolute tolerance here to 1e-6 compared with the default more strict absolute tolerance of 1e-8. https://numpy.org/doc/2.1/reference/generated/numpy.allclose.html

hrajagers (changes requested):
Two comments added. Please update.

tests/batch/test_batch_core.py
MAfarrag: changed the tolerance to 1e-8

MAfarrag:
waiting for https://github.com/Deltares/D-FAST_Commons/pull/1 to be merged.

